PR #17, opened 2026-06-12
[SCU-1491] pi-capabilities: support skills under the pi harness

DanverImbue:
## Summary

Enable **skills** for the `pi` harness. A pi workspace now discovers and runs the same skill set a Claude workspace shows — the slash picker and the skills panel list the workspace's skills, and invoking one causes pi to follow that skill. The `supports_skills` capability flips `True` for pi.

This is Sculptor-side only; the pi binary is not modified.

## How it works

- **`web/skills.py`** gains `get_skill_source_directories()` — the single source of truth for *where* skills live (plugin `skills/`, repo `.claude/skills`, repo `.claude/commands`, and the user-global equivalents under `~/.claude`). `discover_skills` (the authority behind the `/api/v1/skills` picker endpoint) is refactored onto it so the picker list and pi's loaded set stay in lockstep. Discovery behavior is unchanged (covered by the existing tests).
- **`PiAgent`** points pi at those directories with repeatable `--skill` launch flags:
  - `SKILL.md`-directory skills map onto pi's agentskills.io discovery directly.
  - Loose `.claude/commands/*.md` files are not a shape pi discovers, so each is wrapped in a synthesized `SKILL.md` directory under the per-task state dir, and that directory is passed via `--skill`. Because it lives outside `.claude`/`~/.claude`, the picker never lists it twice and a Claude workspace never sees it.
- **Invoking a skill** stays harness-agnostic on the frontend: a picked skill arrives as the same `/name [args]` text Claude receives. `PiAgent` rewrites a leading `/name` into pi's `/skill:<name>` form when `name` is a discovered skill — pseudo-skills (`/clear`, `/copy`, `/btw`) and ordinary slash text are left untouched.

## Known divergences

- **Plugin skills are un-namespaced in pi.** The picker shows a plugin skill as `sculptor-workflow:fix-bug`; pi registers the bare `skill:fix-bug`. The invocation rewrite strips the `<plugin>:` prefix so picking it works. This is fine while no two picker entries share a bare name; giving pi per-namespace skill identities is a sensible follow-up.
- **pi adds its own skill sources.** pi also auto-discovers its native skill locations (`~/.agents/skills`, `~/.pi/agent/skills`, …). Those extra skills appearing under pi is acceptable surplus, not a parity loss.

## Testing

- `just check` (lint, typecheck, ratchets, file-hygiene) — green
- `just test-unit` (backend + frontend) — green
- Integration (fake-pi) `test_pi_capability_gating.py` — the skills panel and slash picker list a seeded workspace skill under pi (18 passed)
- `just test-real-pi` — green: `real_pi/test_skills.py` seeds a sentinel skill, invokes it through the picker path, and the real pi model follows it; the basic-flow, file-edit, and session-resume real-pi tests pass too. (The skills test needed one rerun: an unrelated, pre-existing flake in pi-version parsing — pi's `--version` stderr can interleave a Bun CPU-feature warning whose embedded version is picked up instead of pi's — intermittently raises a version mismatch before the agent starts. Worth a follow-up to anchor the version parse on pi's own line; out of scope here.)
- `just test-real-claude` — could not run in this environment: it aborts at workspace-onboarding setup (`OnboardingWizard is showing instead of the expected app element`), a pre-existing test-harness env prerequisite unrelated to this change (it never reaches Claude). The only Claude-facing code here is the behavior-preserving `discover_skills` refactor, covered by its unit tests, the frontend unit tests, and the Claude branch of the gating integration test.

(Sent by Claude)

### DIFF
--- a/sculptor/frontend/src/components/TipTapConfig.ts
+++ b/sculptor/frontend/src/components/TipTapConfig.ts
@@ -338,7 +338,11 @@ export const createTipTapExtensions = ({
             suggestions: [
               // CAPABILITY-GAP: supportsFileReferences — the @-mention file/folder picker resolves path references the agent reads itself; both Claude and pi report true today, so no harness suppresses it yet. createTipTapExtensions has no taskID, so gate here (thread the capability through Editor) when a harness reports !supportsFileReferences.
               ...(workspaceID && projectID ? [createFileSuggestion(projectID, workspaceID)] : []),
-              // CAPABILITY-GAP: supportsSkills — slash-command skill picker fetches the full skill list regardless of harness; pi has no skills layer, so the items should be suppressed at the picker source when the active task's harness reports !supportsSkills.
+              // The slash-command skill picker fetches the full discover_skills
+              // list regardless of harness, and stays harness-agnostic by
+              // design: pi now supports skills (supports_skills=True), and
+              // PiAgent rewrites a picked `/name` into pi's `/skill:<name>`
+              // form, so the same picker text works for both harnesses.
               ...(workspaceID
                 ? [createSkillSuggestion({ workspaceID })]
                 : projectID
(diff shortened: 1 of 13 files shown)


danverbraganza (approved):
Of course I know him, he's me.

PR #23, opened 2026-06-12
Eliminate imbue_core: fold into sculptor.foundation, then reorganize

saeedn:
## Summary

`imbue_core` had become a sculptor-only dependency rather than the shared library it was originally meant to be. This **eliminates it**: folds its code into `sculptor`, drops the separate-package machinery, and reorganizes so the resulting `sculptor.foundation` holds only leaf utilities.

### 1. Eliminate `imbue_core` → `sculptor.foundation`
- Deleted verified-dead code: `cattrs_serialization` (only self-tested), the `ComputingEnvironment` protocol, and an orphaned `LocalGitRepo` in `git.py` (sculptor already had its own copies under `sculptor/testing/`).
- Moved `imbue_core/imbue_core/*` → `sculptor/sculptor/foundation/`; rewrote ~760 imports.
- Collapsed the uv workspace: dropped the member/source/dependency, folded the still-used third-party deps into `sculptor/pyproject.toml`, regenerated `uv.lock`.
- Rewired ratchets, the Dockerfile copy, and the pyre/ruff/CI config.

### 2. Reorganize `sculptor.foundation` + dead-code scrub
- Moved domain modules to proper homes: `state/` → `sculptor/state/`, telemetry → `sculptor/telemetry/`, `user_config`/`custom_actions` → `sculptor/config/`, both ID modules merged into `sculptor/primitives/ids.py`, `AgentToolName` → `interfaces/agents/`, testing-only modules → `sculptor/testing/`.
- Removed genuinely-dead code (`async_utils`, `get_temp_dir`) via production-reachability analysis.
- Result: `foundation/` is now leaf utilities only.

## Verification
`just format`, `just check` (lint, pyre, ratchets, file-hygiene, uv-lock), and the backend + foundation + sculpt unit suites pass. The only `test-unit` failures are the pre-existing flaky PTY tests, which pass in isolation.

(Sent by Claude)


trisiak (approved):
Workflow changes look fine and benign.

PR #27, opened 2026-06-12
Frontend plugin system scaffolding (SCU-1479)

trisiak:
Closes SCU-1479.

## Summary

Base revision of a **frontend plugin system** for Sculptor — infrastructure
only. No plugin ships bundled, and the whole system sits behind an
experimental flag that is **off by default**, so merging this has zero
visible UX impact. The first real plugins follow as separate PRs.

## What ships

**Experimental flag (the merge gate)**
- Settings → Experimental → "Frontend plugins" (persisted in UserConfig,
  default off). With the flag off, the plugin loader skips bootstrap
  entirely and the Plugins settings section is absent from the settings
  sidebar and the Cmd+K palette. Toggling takes effect after an app reload.

**Runtime plugin loading**
- A plugin is an independent Vite project that builds an ESM bundle served
  from `public/plugins/<id>/` with a `manifest.json` (id, version, entry,
  SDK semver, peer versions).
- A native browser **import map** in `index.html` resolves bare specifiers
  (`react`, `react/jsx-runtime`, `jotai`, `@radix-ui/themes`,
  `@tanstack/react-query`, `lucide-react`, `@sculptor/plugin-sdk`) to
  `/plugin-runtime/` stubs that re-export the host's singleton instances —
  so hooks and context work across the plugin boundary.
- `pluginManager` owns the lifecycle: manifest fetch + SDK-major validation,
  dynamic import, `activate(api)`, unload/reload (cache-busted re-import),
  plus user-managed plugin sources persisted to localStorage.
- Every contributed component is wrapped in a per-plugin error boundary and
  context providers (plugin id, workspace id); a crashing plugin cannot take
  down the host.

**SDK v1 (`@sculptor/plugin-sdk`)**
- Hooks: `useWorkspaceTasks`, `useWorkspaceId`, `useWorkspaceBranch`,
  `usePluginSetting(key)`.
- Host API: `registerPanel` (merges into the existing panel registry, with a
  "plugin" badge in the Panels settings) and `registerSettings` (per-plugin
  settings UI under the plugin's row).

**Shared TanStack Query access**
- Plugins use `@tanstack/react-query` directly against the host's shared
  QueryClient. Scoping is by key namespace: host keys start with the
  reserved `"sculptor"` prefix, plugin keys with the plugin id. A dev-mode
  guard warns on keys outside any namespace. The runtime stub deliberately
  does not export `QueryClient`/`QueryClientProvider`.

## Not in this PR

- No bundled plugin (`BUILTIN_SOURCES` is empty); an example plugin (a
  Linear issue panel) follows in a separate PR.
- The SDK is delivered via the import-map runtime stub, not yet a published
  npm package.
- `usePluginSetting` persists to localStorage; it is not a secret store.

## Docs

`sculptor/frontend/plugins/README.md` documents the loading pipeline, the
SDK surface, and the shared-QueryClient rules (key namespace, explicit
`staleTime`, namespace-scoped invalidation, no secrets in keys).

## Test plan

- `just format` / `just check` green; frontend, imbue-core, and sculpt unit
  suites green (settings-section drift tests updated to cover the gated
  Plugins section in both flag states).
- Manual: with the flag off, settings show no Plugins entry and no plugin
  code runs; with the flag on (after reload), the Plugins section appears
  and a dev plugin source can be added/reloaded/removed end-to-end.

(Sent by Claude)


sculptor/tests/integration/frontend/test_plugin_loader.py
```diff
@@ -0,0 +1,205 @@
...
+    settings_page = navigate_to_settings_page(page=sculptor_instance_.page)
+    settings_page.click_on_experimental().set_frontend_plugins(enabled=True)
+    try:
+        with spawn_plugin_fixture_server() as server:
+            plugins = settings_page.click_on_plugins()
+            _exercise_error_modes(plugins, server)
+            _exercise_valid_load_and_remove(plugins, server)
+    finally:
+        with contextlib.suppress(Exception):
+            settings_page.click_on_experimental().set_frontend_plugins(enabled=False)
```
trisiak: If we fail to disable the plugins, we may leave the test instance in a bad state for future tests. This should just fail on exception, worst case we get double exception.

PR #33, opened 2026-06-13
Add terminal agents: plain shells and registered TUI programs

saeedn:
## Summary

Adds **terminal agents** — a third kind of agent alongside the natively-parsed
Claude and pi chat harnesses. A terminal agent's main panel is a persistent
terminal (PTY) instead of a chat: you run any program (typically an agent TUI)
in a shell inside the workspace's code directory, and Sculptor never parses its
output.

Two flavors ship:

- **Plain Terminal** — a bare login shell in the workspace code directory.
- **Registered terminal agents** — a predetermined program (e.g. the Claude
  Code TUI) launched in that shell, with optional rich integration: coarse
  status (busy / idle / waiting), file-change-driven diff refresh, and
  session-id resume after a restart, all reported to a local HTTP signal API.

A bundled **Claude Code** registration is installed out of the box (as a
user-owned, editable, deletable registration) and doubles as the reference
example for writing third-party registrations.

Along the way, the deliberately-temporary workspace-bound harness selection is
replaced by a **per-agent** type choice made at agent-creation time. The `+`
button becomes a split button: a plain click creates the last-used type, and
the chevron opens a type menu (Claude / pi when enabled / Terminal / each
registered agent).

## Why

Lets users run any agent or program they want inside a Sculptor workspace —
including agent TUIs Sculptor doesn't natively parse — while still getting the
shared diff panel, status dots, branch info, and restart resume. When an agent
ships a feature Sculptor can't yet visualize, its TUI is a full-fidelity
fallback.

## What's included

- Per-agent type model + an agent-type → config factory; removal of the
  workspace-bound harness column (with a schema migration and symmetric
  downgrade).
- A dedicated terminal-agent task handler owning an agent-scoped PTY
  (launch / resume / periodic diff refresh / teardown), registered under the
  workspace environment id so environment teardown is the cleanup backstop.
- A local HTTP signal API + `sculpt signal` CLI subcommands
  (busy / idle / waiting / files-changed / session-id); run-scoped status that
  drives the same tab indicators native agents use and resets across a backend
  restart.
- A TOML registration loader (re-read on menu open), session-id charset
  validation plus shell-quoting for the resume template, and stale-shell
  reaping on restart.
- A server-side automated-prompt reverse channel so Commit / Create PR / custom
  actions can drive an opt-in registered terminal agent (state-gated, wrapped
  in bracketed paste).
- The bundled Claude Code registration and its hooks.

## Testing

- `just format` / `just check` (lint, typecheck, ratchets, generated types) /
  `just test-unit` (backend, frontend, imbue-core, sculpt): all green, no
  generated-file drift.
- Feature Playwright integration suite: green (16 tests across the feature
  files), plus a real-Claude end-to-end test for the bundled registration.
- ~110 new backend unit and integration tests covering the config factory, the
  status driver (including the run-scoped reset across a restart), the signal
  API and its validation, lifecycle (launch / resume / reaping / teardown), and
  the UI flows.

## Review

This branch went through the spec → architecture → plan → build → review
workflow. The final review (`agent_docs/terminal-agents/review.md`) confirmed
all requirements are covered and resolved five medium findings: a
bundled-registration security note in the README, a flaky integration test, a
stale code comment, and two registered-agent resilience edge cases on the
agent-creation surfaces.

(Sent by Claude)


samples/terminal_agents/claude-code/claude-code-hooks.json
```diff
@@ -0,0 +1,81 @@
...
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "sculpt signal busy || true"
+          },
+          {
+            "type": "command",
+            "command": "sid=$(sed -n 's/^{[[:space:]]*\"session_id\"[[:space:]]*:[[:space:]]*\"\\([^\"]*\\)\".*/\\1/p'); [ -n \"$sid\" ] && sculpt signal session-id \"$sid\" || true"
```
saeedn: Future enhancement: this can be made more robust if we package and/or download `jq`

samples/terminal_agents/claude-code/claude-code-hooks.json
```diff
@@ -0,0 +1,81 @@
...
+          {
+            "type": "command",
+            "command": "sculpt signal waiting || true"
+          }
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": "Edit|Write|MultiEdit|NotebookEdit|Bash",
```
saeedn, line 62: Future enhancement: replace this self-reporting of file modifications with a file watcher

sculptor/builder/build-sidecar.sh
```diff
@@ -64,6 +64,7 @@ pyinstaller --onedir --name sculptor_backend \
   --add-data "sculptor-plugin:sculptor-plugin" \
   --add-data "sculptor-workflow:sculptor-workflow" \
   --add-data "sculptor-experimental:sculptor-experimental" \
+  --add-data "../samples/terminal_agents:samples/terminal_agents" \
```
saeedn, line 67: Need to verify this works with a packaged build

sculptor/frontend/src/pages/workspace/components/useTerminalChatActions.ts
```diff
@@ -0,0 +1,67 @@
...
+ * single seam that makes the prompt-driven features (Commit, Create PR,
+ * custom actions) work for automated-prompt-capable terminal agents.
+ *
+ * When the agent's registration opted in (`acceptsAutomatedPrompts`),
+ * `sendMessage` routes the prompt through the terminal-input endpoint so it
+ * arrives as typed input; otherwise nothing is registered and every consumer
+ * stays disabled by default. Consumers are untouched — the routing decision
+ * lives entirely in which hook registered the actions.
+ */
+export const useTerminalChatActions = (taskId: string): void => {
```
saeedn, line 20: Bug (fix separately): This works correctly for actions that send immediately, but is a noop for actions that do not auto-send. Determine if it's possible to preserve that behavior. If not, disable non-auto-send actions when a terminal agent is active.

sculptor/sculptor/services/ci_babysitter_service/coordinator.py
```diff
@@ -350,15 +350,52 @@ def _select_model_for_workspace(
...
+    def _select_chat_agent_config_for_workspace(
+        self,
+        workspace_id: WorkspaceID,
+        project_id: ProjectID,
+        transaction: DataModelTransaction,
+    ) -> ClaudeCodeSDKAgentConfig | PiAgentConfig:
+        """Inherit the chat-agent config type from the workspace's most recent
+        chat agent, falling back to Claude.
+
+        Terminal agents are skipped — the babysitter is always a chat agent
```
saeedn, line 364: Bug (fix separately): if the MRU agent is a terminal agent that supports prompts, it should be selected. We should always preserve the MRU agent type when possible.

sculptor/sculptor/services/terminal_agent_registry/bundled.py
```diff
@@ -0,0 +1,89 @@
...
+
+from sculptor.common.plugin import get_plugins_base_dir
+from sculptor.services.terminal_agent_registry.registry import get_registrations_dir
+
+_SENTINEL_FILE_NAME = ".claude-code.installed"
+_BUNDLED_FILE_NAMES = ("claude-code.toml", "claude-code-hooks.json")
+_HOOKS_FILE_NAME = "claude-code-hooks.json"
+# The sample TOML references its hooks file at the manual-install location;
+# the installer rewrites it to wherever the hooks actually land.
+_SAMPLE_HOOKS_PATH_TOKEN = "~/.sculptor/terminal_agents/claude-code-hooks.json"
```
saeedn: This seems brittle. Should we instead have a replaceable token like `{sculptor_directory}`?

sculptor/sculptor/services/terminal_agent_registry/bundled.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def get_bundled_claude_code_dir() -> Path | None:
+    """Locate the shipped claude-code sample directory, or None if absent.
+
+    Packaged app: bundled as data next to the plugins (`_internal/samples/…`).
+    Running from source: `samples/` at the repo root.
+    """
+    base = get_plugins_base_dir()
+    candidates = (
```
saeedn, line 41: Why do we need to have multiple candidate paths? Are they different between dev and prod?

sculptor/sculptor/tasks/handlers/run_terminal_agent/diff_refresh.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _compute_fingerprint(self) -> str | None:
+        """Hash of `git status --porcelain` + HEAD, or None on git failure.
+
+        A transient git failure (index lock, repo mid-rewrite) must not kill
+        the agent loop — swallow and retry on the next interval.
+        """
+        try:
+            status = run_blocking(
+                command=["git", "status", "--porcelain"],
```
saeedn, line 69: Verify: how expensive is this to do in a polling loop? And how urgent is the follow-up work to add a file watcher?

sculptor/sculptor/tasks/handlers/run_terminal_agent/v1.py
```diff
@@ -0,0 +1,278 @@
...
+    # SCULPT_AGENT_ID so `sculpt signal …` can identify this agent. The pty
+    # scrubs inherited SCULPT_*/SCULPTOR_* vars and re-applies extra_env (PATH
+    # is prepended), so everything must go through extra_env.
+    #
+    # SCULPT_PLUGINS_DIR and SCULPT_CLAUDE_BIN exist for registration launch
+    # commands (shell-expanded at launch): install-relative paths move on app
+    # updates — and on every AppImage launch — so a registration file cannot
+    # bake them in.
+    extra_env: dict[str, str] = {
+        "SCULPT_API_PORT": str(settings.BACKEND_PORT),
```
saeedn: Is this list of env vars duplicated with the other agent run loop? Should we have a single unified list of pass-through env vars so they always match?

sculptor/sculptor/web/app.py
```diff
@@ -527,12 +546,17 @@ def start_task(
                 )
                 logger.debug("Created workspace {} for task {}", workspace.object_id, task_id)
 
-            # Auto-assign "Agent N" name when no explicit name is provided
+            # Prompt-ful creation is always a chat agent — terminal agents have
```
saeedn, line 549: Follow-up work: ensure that the `sculpt` commands for creating/running agents work correctly and fix their ergonomics as needed

sculptor/builder/build-sidecar.sh
saeedn: Done. Verified it works correctly with `just refresh app`

samples/terminal_agents/claude-code/claude-code-hooks.json
saeedn: Hardened the `sed` extraction in 700c393 — it now matches the `session_id` key anywhere in the payload (no longer anchored to the first key on the first line) and takes the first hit, so it survives compact or pretty-printed JSON and key reordering. Bundling/downloading `jq` for a fully robust parse is captured as a deferred work item to file after this PR merges.

(Sent by Claude)

samples/terminal_agents/claude-code/claude-code-hooks.json
saeedn: Deferred to a post-merge work item (tracked together with the diff-refresh file-watcher below). Per your clarification, the scope is narrow: a watcher would replace the self-reported file-modification signal (`files-changed`) and the git-status polling — not the busy/idle/waiting signals, which a filesystem watcher can't infer. No change in this PR.

(Sent by Claude)

sculptor/builder/build-sidecar.sh
saeedn: Verified against a packaged build via `just refresh app` — the bundled `samples/terminal_agents` data resolves correctly. Resolved.

(Sent by Claude)

sculptor/frontend/src/pages/workspace/components/useTerminalChatActions.ts
saeedn: Confirmed; deferred to a bug filed after merge. Current leaning is to disable the non-auto-send actions for terminal agents (staging text into an editor that doesn't exist has no sensible terminal analog), but the bug can make that call. No change in this PR.

(Sent by Claude)

sculptor/sculptor/services/ci_babysitter_service/coordinator.py
saeedn: Deferred to a bug filed after merge — preserve the MRU agent type when the most-recent agent is a prompt-accepting terminal agent, rather than always falling back to Claude. No change in this PR.

(Sent by Claude)

sculptor/sculptor/services/terminal_agent_registry/bundled.py
saeedn: Done in 9f931e3. Replaced the install-time path rewrite with proper render-time template variables: registrations can use `{sculptor_directory}` and `{terminal_agents_directory}` (alongside the existing `{session_id}`), substituted to absolute paths when the launch/resume command is rendered — for every registration, so the variables also work in hand-written registrations, not just the bundled sample. The loader now validates both `launch_command` and `resume_command_template` against this known vocabulary and rejects any unknown `{...}` token (a typo fails loudly at load). The bundled sample uses `{terminal_agents_directory}/claude-code-hooks.json`, and the installer now copies files verbatim (no rewrite).

(Sent by Claude)

sculptor/sculptor/services/terminal_agent_registry/bundled.py
saeedn: Yes — they differ between dev and prod: a packaged build resolves to `_internal/samples/...` (`base / "samples"`), and a source checkout to the repo-root `samples/` (`base.parent / "samples"`, since `get_plugins_base_dir()` is the `sculptor/` project dir there). Added a code comment in 9d7cc4d spelling out both layouts.

(Sent by Claude)

sculptor/sculptor/tasks/handlers/run_terminal_agent/diff_refresh.py
saeedn: Cost: each tick is `git status --porcelain` + HEAD, hashed, self-limited to once per 3s per terminal agent with a 5s timeout — negligible on a normal repo (tens of ms); it only becomes notable on very large / monorepo working trees. Correctness is bounded and failure-swallowing. So the file watcher is a performance/scale follow-up, not urgent — filed as a post-merge work item (which also subsumes the `files-changed` hook self-report from the comment above). No change in this PR.

(Sent by Claude)

sculptor/sculptor/tasks/handlers/run_terminal_agent/v1.py
saeedn: Yes — it was duplicated across three sites (this handler, the chat handler, and the workspace terminal manager) and had already drifted. Unified behind a single `build_sculpt_backend_env` helper in 81fccb8 so the `SCULPT_*` identity vars can't diverge. `PATH` stays per-site (each computes it differently) and workspace terminals omit the agent id since they aren't agent-scoped.

(Sent by Claude)

sculptor/sculptor/web/app.py
saeedn: Deferred to a post-merge work item — I'll turn it into concrete items (which `sculpt` subcommands, and what specifically to fix) when filing. No change in this PR.

(Sent by Claude)

sculptor/frontend/src/common/state/atoms/userConfig.ts
```diff
@@ -200,11 +200,11 @@ export const isRichMarkdownRenderingEnabledAtom = atom<boolean>(
   (get) => get(userConfigAtom)?.enableRichMarkdownRendering ?? false,
 );
 
-// Multi-harness / pi (experimental — off by default). Gates only the
+// Pi agent (experimental — off by default). Gates only the
 // new-workspace harness picker and the harness badge; the workspace.harness
 // execution path is independent of this flag, so existing pi workspaces run pi
 // regardless.
-export const isMultiHarnessEnabledAtom = atom<boolean>((get) => get(userConfigAtom)?.enableMultiHarness ?? false);
+export const isPiAgentEnabledAtom = atom<boolean>((get) => get(userConfigAtom)?.enablePiAgent ?? false);
```
DanverImbue, line 207: Mm, no, I did actually mean "isMultiHarnessEnabledAtom" for this--because I was envisaging this being the extension point for other Harnesses like OpenCode etc.

saeedn: I saw that that was the original intent, but this branch already invalidates it since we're adding terminal agents without a feature flag. So this is now a misnomer since "multi-harness" is now enabled out of the box.

If we do add other harnesses in the future, I think they could just get their own per-harness feature flags.

DanverImbue: Not a blocker, though. I can always change this back when we add the next harness.

PR #46, opened 2026-06-16
Make migrations immune to triggers referencing dropped columns

saeedn:
## Problem

Auto-managed tables carry triggers (`<table>_before_insert`, `set_<table>_created_at`) whose bodies reference *every* column of the table. SQLite validates trigger bodies during `ALTER TABLE`, so any migration that drops a column a trigger still references fails at startup:

```
error in trigger workspace_before_insert after drop column: no such column: NEW.harness
```

Until now, every column-dropping migration had to *remember* to drop its triggers first. That convention is forgettable, and the data shows it: while the drop-on-upgrade migrations handle it, ~10 add-column migrations' **downgrade** paths still drop their column with no trigger drop — latent copies of the same bug. It has broken startup more than once.

## Fix

Replace the convention with an enforced **invariant: no auto-managed trigger exists while migrations run.**

- New `drop_all_automanaged_triggers()` helper.
- Called at the single migration chokepoint, which is present in both runner paths — `env.py` (file databases) and `override_run_env` (in-memory) — immediately before migrations execute.
- Triggers are recreated from the current model right after migrations by the existing `initialize_db_from_connection()`, so the end state is unchanged.

This makes the failure structurally impossible rather than policed: no migration — upgrade or downgrade, now or future — ever needs to drop triggers itself. It also makes the upgrade path consistent with the fresh-install path, which migrations already assume (one migration maintains a `_latest` table by hand "because triggers don't exist during migration" — only true on a fresh install until now).

The existing per-migration `DROP TRIGGER` blocks are now redundant belt-and-suspenders and are intentionally left untouched, so the standalone bugfix they belong to remains valid on its own.

## Tests

Three guard tests (all passing):
1. The exact failure mode — a trigger referencing a column blocks `DROP COLUMN`, and the helper unblocks it.
2. & 3. End-to-end: each runner path (in-memory and file) drops pre-existing triggers before migrating. Verified that test no. 2 **fails** if the wiring is removed.

`just format`, `just check`, and `just test-unit` all pass.

(Sent by Claude)

### DIFF
--- a/sculptor/sculptor/database/alembic/env.py
+++ b/sculptor/sculptor/database/alembic/env.py
@@ -4,6 +4,7 @@ from alembic import context
 from sqlalchemy import engine_from_config
 from sqlalchemy import pool
 
+from sculptor.database.alembic.utils import drop_all_automanaged_triggers
 from sculptor.database.core import METADATA
 from sculptor.services.data_model_service.sql_implementation import register_all_tables
 
@@ -75,6 +76,10 @@ def run_migrations_online() -> None:
         context.configure(connection=connection, target_metadata=target_metadata)
 
         with context.begin_transaction():
+            # Invariant: no auto-managed triggers exist while migrations run, so a migration
+            # can never fail by dropping a column a trigger still references. Triggers are
+            # recreated from the current model immediately after migrations.
+            drop_all_automanaged_triggers(connection)
             context.run_migrations()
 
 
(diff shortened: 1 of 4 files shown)


trisiak (approved):
The triggers during migrations may propagate to the "latest" snapshot. With this change, keeping things in sync is a responsibility of the migration script.

That being said, I'm pretty sure the non-latest entries are not being ever read and are a data debt that is write-only. So if there's something to change, it would be to abandon the snapshots entirely -- way out of scope here.

PR #58, opened 2026-06-17
Serve the packaged renderer from a custom secure app:// origin

trisiak:
**Draft / exploration.** Opened to get CI coverage (the test workflow only runs on PRs). This is the focused "custom protocol" prototype that the runtime-plugin work will build on later — it deliberately contains **only** the protocol/origin change, no plugin code.

## Why

The packaged Electron app loads its renderer from `file://`, which gives the page the opaque `"null"` origin. That breaks the web primitives a normal page relies on:

- absolute path resolution points at the filesystem root, not the app bundle
- `fetch()` of local resources is blocked
- dynamic `import()` is restricted
- there's no real security context for CSP

It's also the blocker for runtime-loaded frontend plugins, which need a stable origin so their import map and same-origin loading behave predictably.

## What changed

Introduce a custom `sculptor://app` scheme, registered as **standard + secure + fetch-capable**, and serve the built renderer bundle through it via `protocol.handle` instead of `file://`:

- **`appProtocol.ts`** — the URL→file mapping and a path-traversal guard as pure, Electron-free helpers (so they're unit-testable without an Electron runtime). Traversal is contained by two layers: the standard scheme normalizes `../` at the origin root, and a `path.resolve` + prefix check is the defense-in-depth net.
- **`main.ts`** — registers the scheme before app-ready, registers the file handler on ready (with an extensionless-path SPA fallback), and points the production renderer URL at `sculptor://app/index.html`. **Development is unchanged** (still the Vite dev server over http).
- **`web/app.py`** — the backend CORS allowlist accepts the new origin.

The `app` host prefix is intentionally a single origin so later work can serve plugins and their runtime stubs under the same origin **by path** (`/plugins/<id>/...`, `/plugin-runtime/...`) with no cross-origin plumbing.

## Test coverage / gap

- `appProtocol.test.ts` covers the path mapping, traversal containment, scheme/host rejection, and SPA fallback.
- **Gap:** the existing Electron integration tests run against the dev server (http), so the live `protocol.handle` path isn't exercised end-to-end yet. A packaged-mode integration test is a sensible follow-up.

(Sent by Claude)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LeQKimzSyEDjr9RpgX8w9o)_

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,8 +9,14 @@
 
 FROM ubuntu:24.04
 
+# xvfb + xauth provide the virtual X display the Electron integration launch
+# mode needs (the harness wraps Electron in `xvfb-run` when there's no DISPLAY).
+# Electron's other shared libs come from `playwright install --with-deps
+# chromium` in finalize-base-image.sh, which also handles Ubuntu 24.04's t64
+# package renames.
 RUN apt-get update && apt-get install -y --no-install-recommends \
         git curl ca-certificates build-essential pkg-config libssl-dev \
+        xvfb xauth \
     && rm -rf /var/lib/apt/lists/*
 
 # Install uv
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -11,12 +11,16 @@ FROM ubuntu:24.04
 
 # xvfb + xauth provide the virtual X display the Electron integration launch
 # mode needs (the harness wraps Electron in `xvfb-run` when there's no DISPLAY).
-# Electron's other shared libs come from `playwright install --with-deps
-# chromium` in finalize-base-image.sh, which also handles Ubuntu 24.04's t64
-# package renames.
+# libgtk-3-0t64 + libnotify4 are the GTK GUI stack that full Electron links
+# against: `playwright install --with-deps chromium` (finalize-base-image.sh)
+# installs deps for Playwright's *headless* Chromium shell, which omits GTK — so
+# without these the Electron binary aborts at startup with "libgtk-3.so.0:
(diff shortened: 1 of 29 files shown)


trisiak:
Electron lane failure is a GPU/GL env issue, not the `sculptor://app` change. Every failing sandbox logs `No suitable EGL configs found` → `Exiting GPU process due to errors during initialization`, so the renderer never paints and `_reset_browser_state`/`expect_app_not_onboarding` times out. Clean correlation in the artifact: 136/136 failure logs have the EGL error, 0/6 passing logs do. The app does load over the new origin in CI (`Initial URL: sculptor://app/index.html`, `SPA initial render: 5.27s`), and the exact commit passes locally in electron mode where GL works.

Cause: the image installs `xvfb xauth libgtk-3-0t64 libnotify4` but no Mesa GL driver, so full Electron has no EGL config to bind (Playwright's bundled-ANGLE headless shell doesn't need one). Fix: add `libgl1-mesa-dri` to the Dockerfile apt install (pulls llvmpipe/`swrast_dri.so`), and/or pass software-GL flags to `electron:start` (`--use-gl=angle --use-angle=swiftshader`, or `--disable-gpu`).

(Sent by Claude)

---
_Generated by [Claude Code](https://claude.ai/code)_

trisiak:
Superseded — this PR is being split into two stacked PRs for cleaner review and independent CI signal:

1. **#62** — [SCU-1516] Run the @electron integration tests in the CI offload lane (targets `main`)
2. **#63** — [SCU-1515] Serve the packaged renderer from a custom secure `sculptor://app` origin (stacked on #62)

The combined diff of #62 + #63 is byte-identical to this PR. Closing this in favor of the stack.

(Sent by Claude)

PR #62, opened 2026-06-17
[SCU-1516] Run the @electron integration tests in the CI offload lane

trisiak:
**Base of a 2-PR stack split out of #58.** Targets `main`. The follow-up PR (SCU-1515, the `sculptor://app` custom protocol, #63) is stacked on top of this branch.

## Why

The offload CI setup only runs the `browser` launch mode. The `@electron`-marked integration tests (real Electron driven over CDP) are skipped in `browser` mode, so after the move to the GitHub/offload setup that ~36-test subset runs in **no CI lane at all** — a coverage gap introduced by the migration. This PR closes that gap; it's post-cleanup for the move-to-GitHub effort.

## What changed

- **`offload-electron.toml` + `offload.yml` job + `just test-offload-electron`** — a parallel offload lane that runs exactly the `-m electron` subset with `--sculptor-launch-mode=electron`, with its own results dir so it doesn't clobber the browser run. Left additive: the existing browser lane is untouched.
- **`.devcontainer/Dockerfile`** — install `xvfb`/`xauth` (the virtual display the Electron harness wraps the app in), the GTK runtime libs full Electron links against (`libgtk-3`, `libnotify`; Playwright's headless-Chromium dep set deliberately omits the GTK GUI stack), and put the nvm node/npm on `PATH` for non-interactive shells (the harness subprocess doesn't source nvm).
- **Headless launch hardening** — pass `--disable-gpu` on the xvfb path so Chromium uses software rendering directly instead of relying on the crash-looping GPU-process fallback.
- **Launch/CDP failure diagnostics** (`electron_frontend.py`, `subprocess_utils.py`) — capture a bounded tail of Electron's output and include it (plus exit code, and whether `DevTools listening on ws://…` ever printed) in the errors raised on the launch-failure and CDP-setup paths. Offload only streams the failure summary, not per-sandbox logs, so without this the real cause is invisible in CI.
- **Origin/harness adaptations needed for the `@electron` subset to pass** — in Electron mode the renderer is served from its own origin, distinct from the backend's HTTP API origin (the Vite dev server on a separate port in a dev/test run; `sculptor://app` once the protocol PR lands). So:
  - between-test resets navigate to the SPA's own origin (`frontend_url`, captured from the page) instead of the backend URL — the backend serves no SPA, so the old reset left a blank page and **every `@electron` test failed in setup**. `base_url` is renamed to `backend_api_url` to make the backend-API origin explicit.
  - integration tests that built their backend API URL from `page.url` now use the fixture's `backend_api_url` (Playwright's `APIRequestContext` rejects a non-`http` page origin). No behavior change in browser mode.
- **Smaller test fixes for the subset under offload parallelism** — longer webview load timeouts for the browser-panel waits; fix `AutoUpdateTestServer.set_offline` writing the wrong attribute so the offline simulation actually returns 503.

## Note

This is the first run that exercises the real Electron path in CI; treat the initial runs as validation of whether Electron launches under xvfb in a Modal sandbox.

## Relation to #58

#58 bundled this offload lane together with the `sculptor://app` custom-protocol change and was dead-ended in favor of this 2-PR stack: this one (the offload lane + the harness adaptations that make the `@electron` subset green) first, then SCU-1515 (the protocol itself, #63) on top. The combined diff of the two PRs is line-for-line identical to #58.

(Sent by Claude)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -11,13 +11,10 @@ FROM ubuntu:24.04
 
 # xvfb + xauth provide the virtual X display the Electron integration launch
 # mode needs (the harness wraps Electron in `xvfb-run` when there's no DISPLAY).
-# libgtk-3-0t64 + libnotify4 are the GTK GUI stack that full Electron links
-# against: `playwright install --with-deps chromium` (finalize-base-image.sh)
-# installs deps for Playwright's *headless* Chromium shell, which omits GTK — so
-# without these the Electron binary aborts at startup with "libgtk-3.so.0:
-# cannot open shared object file" and never opens its remote-debugging port.
-# (t64 is Ubuntu 24.04's 64-bit-time_t package rename; libgtk-3-0t64 pulls in
-# the pango/cairo/gdk-pixbuf/atk closure transitively.)
+# libgtk-3-0t64 + libnotify4 are the GTK stack full Electron links against:
+# `playwright install --with-deps chromium` only installs deps for Playwright's
+# headless Chromium, which omits GTK. (t64 is Ubuntu 24.04's time_t package
+# rename; libgtk-3-0t64 pulls in the pango/cairo/gdk-pixbuf/atk closure.)
 RUN apt-get update && apt-get install -y --no-install-recommends \
         git curl ca-certificates build-essential pkg-config libssl-dev \
         xvfb xauth libgtk-3-0t64 libnotify4 \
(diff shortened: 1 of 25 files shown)


DanverImbue (approved):
LGTM.

Some of the AI-generated comments are a bit heavy, you might get some value out of prompting:

> For each code comment, ask: does this help future maintainers understand the code, or merely explain today’s bug fix? Remove incidental history, persuasive rationale, and correctness arguments.

Leaving that up to you.

PR #73, opened 2026-06-17
Generate plugin runtime stubs from module namespaces (SCU-1488)

trisiak:
## What

The frontend plugin system's `/plugin-runtime/*.js` stubs were hand-maintained re-export lists. A plugin importing a name that wasn't enumerated (e.g. an un-listed lucide icon) silently got `undefined` at runtime, with no build error — the most likely papercut as plugins multiply.

This replaces them with **build-time generation**: a Vite plugin (`vite-plugins/plugin-runtime-stubs.ts`) derives each stub's export list from the *actual* installed module namespace, serves the stubs in dev (middleware), and emits them into the build output at the same path. The import map in `index.html` is unchanged.

## Highlights

- **No hand-maintained export lists.** Each stub mirrors exactly what the host package really exports — lucide now exposes all ~5,200 icons as named imports (plus the default-export `Proxy` escape hatch); radix/jotai/tanstack expose their full surface.
- **Byte-stable** across builds with unchanged deps (names are sorted; verified by diffing two builds).
- **Deliberate API boundaries preserved** as small, documented exclusions — e.g. `QueryClient`/`QueryClientProvider` stay withheld so plugins can't construct a second client that would detach from the host's shared cache.
- **Correctness carried over from the old stubs:** `createRoot`/`hydrateRoot` bind to `react-dom/client` (avoiding the deprecation warning), and `jsxDEV` is correctly absent from `react/jsx-runtime`.
- **Host package versions embedded** and exposed via a virtual module and on `window.__SCULPTOR_HOST__.versions`, re-opening the door to enforceable plugin peer-dependency ranges. The loader-side check is left for a follow-up.

## Tests

- `src/plugins/pluginRuntimeStubs.test.ts` covers the pure generation logic: byte-stability, sorting, default vs proxy export, exclusions, multi-source precedence (react-dom), invalid-identifier skipping, and SDK-barrel parsing.
- Validated end-to-end manually: a real `vite build` emits all stubs; two builds diff identical; the dev server serves them with the right content-type; and the host bundle contains the frozen versions object with real versions.

`just format`, `just check`, and the frontend unit tests pass.

SCU-1488

(Sent by Claude)


trisiak:
Thanks — all three addressed in b2c26af:

- **Namespace-reading path now tested.** Extracted `collectStubs`/`collectHostVersions` and added a smoke test that dynamic-imports the *real* installed packages and asserts on the generated stubs (react's `useState` + default export, the tanstack `QueryClient`/`QueryClientProvider` boundary holding against the real namespace, lucide's Proxy + 1000+ enumerated icons, SDK names parsed from the barrel), plus a byte-stability-across-runs check and a real-version check per embedded package.
- **Guard names every missing host key.** Generated stubs now compute `const __missing = [...keys].filter(k => !host || !host[k])` and throw listing all absent keys, instead of only the first source's key.
- **Reserved-word guard.** `isExportableName` now rejects JS reserved words (and strict-mode-illegal binding names like `eval`/`arguments`), so an upstream export name can never emit an unparseable `export const`. Verified by `node --check` over every generated stub.

`just format`, `just check`, and the frontend unit tests pass.

(Sent by Claude)

PR #84, opened 2026-06-18
Add a --harness field to sculpt agent create (SCU-1532)

saeedn:
## Original bug

> `sculpt agent create` has no way to choose the harness/agent type, so it can't create the different kinds of agents. Add a harness field that exposes the **same harness list the UI shows** — "Claude", "Pi", "Terminal", and "Claude CLI" — so the CLI and the UI offer the same choices (how those map to underlying harness types / terminal-agent registrations is an implementation detail to keep out of the interface). When no value is provided, default to the most-recently-used (MRU) harness type rather than a hardcoded default.

Linear: [SCU-1532](https://linear.app/imbue/issue/SCU-1532)

## Hypotheses considered

1. **`sculpt agent create` exposes no harness selection** — it always builds a `CreateAgentRequest` with no `agent_type`, so the server defaults to `claude`. The CLI therefore cannot create Pi, Terminal, or registered ("Claude CLI") agents, and has no MRU default. — **REPRODUCED**.

(The ticket describes a single missing capability, so only one hypothesis was needed; no others were formed or discarded.)

### `sculpt agent create` can only ever create a Claude agent

**Repro steps**

1. From a Sculptor checkout, run the sculpt CLI's `agent create` command and look for a way to pick the harness/agent type: `sculpt agent create --help`. There is no `--harness` (or equivalent) option — only `--workspace`, `--prompt`, `--model`, `--name`, `--json`, `--base-url`.
2. Equivalently, as an automated repro, add a test asserting that `sculpt agent create --harness Pi` sends `agentType: "pi"` to `POST /workspaces/{id}/agents`. Run it: `uv run --project tools/sculpt python -m pytest tools/sculpt/tests/unit/test_agent.py::TestAgentCreateHarness`.
3. Observe that every such test fails: Typer rejects the unknown `--harness` option and exits with code 2.

**Expected vs actual**

- Expected: the CLI offers the same harnesses the UI does — "Claude", "Pi", "Terminal", and registered terminal agents by display name (e.g. "Claude CLI") — maps the choice to the request's `agent_type` / `registration_id`, and defaults to the most-recently-used harness when none is given.
- Actual: there is no harness option at all. Every agent created through the CLI is a Claude agent.

**Before**

The new tests fail because the option does not exist (Typer exits 2 = "No such option: --harness"):

```
$ uv run --project tools/sculpt python -m pytest tools/sculpt/tests/unit/test_agent.py::TestAgentCreateHarness
E       assert 2 == 0
E        +  where 2 = <Result SystemExit(2)>.exit_code
FAILED ...::test_create_with_harness_pi_sends_pi_agent_type
FAILED ...::test_create_with_harness_terminal_sends_terminal_agent_type
FAILED ...::test_create_with_harness_claude_cli_resolves_registered_agent
FAILED ...::test_create_without_harness_defaults_to_most_recently_used
FAILED ...::test_create_with_invalid_harness_errors_and_lists_valid_options
5 failed
```

**Root cause**

`tools/sculpt/sculpt/commands/agent.py`'s `create` command built `CreateAgentRequest(...)` without ever setting `agent_type` or `registration_id`. The generated client already carries both fields (mirroring the backend's `CreateAgentRequest`), and the backend resolves `agent_type` (`claude` / `pi` / `terminal` / `registered` + `registration_id`) into the concrete agent config. The CLI simply never surfaced that choice, so the request always fell back to the server's default of `claude`. The UI, by contrast, builds its chooser from the built-in types plus the registered terminal agents returned by `GET /terminal-agent-registrations` and remembers the last-used choice.

**Fix**

Add a `--harness` option to `sculpt agent create` and a small `sculpt/harness.py` module that mirrors the UI's chooser. Built-in names ("Claude", "Pi", "Terminal") resolve case-insensitively with no server round-trip; any other name is matched against the live registered terminal agents (so "Claude CLI" maps to `agent_type=registered`, `registration_id=claude-code`), and an unknown name errors with the list of valid options. When `--harness` is omitted, the most-recently-used harness is reused — persisted to a small local CLI state file (`~/.sculpt/cli_state.json`, overridable via `SCULPT_STATE_DIR`), encoded the same `registered:<id>` way the frontend stores it — and seeded to Claude on first use rather than a hardcoded default. Because terminal/registered agents cannot take an initial prompt (the server rejects it), the CLI guards `--harness Terminal`/`Claude CLI` + `--prompt` with a clear client-side error. The hardcoded `interface="API"` is unchanged: the server's no-prompt creation path ignores `interface`, so terminal agents are created correctly.

**Test**

- Path: `tools/sculpt/tests/unit/test_agent.py` (`TestAgentCreateHarness`)
- Kind: unit. The testing config requires an integration (Playwright) test unless reproducing through the UI is impossible; the "impossible" criterion that applies here is *"the buggy code path is not reachable from any UI surface (e.g. it is internal-only, a CLI-only entrypoint…)"*. `sculpt agent create` is a CLI-only entrypoint, so a `CliRunner` + `respx` unit test is the correct kind. (Conventions follow the existing `tools/sculpt` CLI unit tests, which use `respx` HTTP mocking — established for testing the CLI without a running server.)
- Failing-test commit: `7eda5d8751`
- Fix commit: `a534fc6f5c`
- Follow-up test (prompt guard, from self-review): `8069a8a8f9`

**After**

The tests pass once the option exists and maps the harness correctly (6 = the original 5 plus the self-review prompt-guard test):

```
$ uv run --project tools/sculpt python -m pytest tools/sculpt/tests/unit/test_agent.py::TestAgentCreateHarness
......                                                                   [100%]
6 passed
```

`sculpt agent create --help` now shows the option:

```
--harness   TEXT  Harness to create: Claude, Pi, Terminal, or a registered terminal
                  agent by name (e.g. 'Claude CLI'). Defaults to the most-recently-used
                  harness.
```

Full pre-commit verification is green: the whole sculpt CLI suite (258 tests), plus `just check` and `just test-unit`.

## Scope

This change covers `sculpt agent create` as the ticket specifies. `sculpt run` (which creates a workspace + agent in one step, and is built around a required prompt) still hardcodes Claude; extending harness selection there is a separate decision — terminal/registered agents reject the prompt that `run` requires — and is intentionally left out.

## Review notes

A self-review pass with `/code-review-checklist` was run against `<merge-base>...HEAD`. The one MEDIUM finding — a missing test for the new `--prompt` + terminal-harness guard — was acted on in commit `8069a8a8f9`. Remaining low-severity notes, left as-is:

- A most-recently-used `registered:<id>` whose registration was deleted is sent to the server on the next bare `sculpt agent create`, which returns a 422. This matches the UI (a plain "+" click sends the stored choice and lets the server reject), so it is left as-is.
- `write_most_recently_used_harness` intentionally swallows `OSError` on the state-file write: persisting the MRU is a convenience cache and must not fail agent creation (e.g. on a read-only home). This is documented at the call site.

(Sent by Claude)

### DIFF
--- a/tools/sculpt/sculpt/harness.py
+++ b/tools/sculpt/sculpt/harness.py
@@ -2,16 +2,12 @@
 
 Mirrors the harness chooser the UI shows: the built-in Claude, Pi, and
 Terminal types plus any registered terminal agents (for example
-"Claude CLI"). The most-recently-used selection is persisted to a small
-local state file so a bare ``sculpt agent create`` reuses it, matching the
-UI's MRU default.
+"Claude CLI"). When no harness is requested the CLI sends nothing and lets
+the server apply the user's most-recently-used choice from the Sculptor
+app, so ``sculpt`` keeps no harness state of its own.
 """
 
-import json
-import os
-from collections.abc import Mapping
 from collections.abc import Sequence
-from pathlib import Path
 
 from pydantic import BaseModel
 
@@ -26,14 +22,6 @@ BUILTIN_HARNESS_LABELS: dict[AgentTypeName, str] = {
     AgentTypeName.TERMINAL: "Terminal",
 }
 
-_STATE_DIR_ENV_VAR = "SCULPT_STATE_DIR"
-_DEFAULT_STATE_DIR = "~/.sculpt"
-_STATE_FILE_NAME = "cli_state.json"
-_MRU_HARNESS_KEY = "mru_harness"
-# Encoding for a registered terminal agent in the MRU file, matching the
-# frontend's ``registered:<id>`` StoredAgentType form.
-_REGISTERED_PREFIX = "registered:"
-
 
 class HarnessSelection(BaseModel):
     """A resolved harness choice.
@@ -43,12 +31,6 @@ class HarnessSelection(BaseModel):
 
     agent_type: AgentTypeName
     registration_id: str | None = None
-
-    def encode(self) -> str:
-        """Encode this selection for the MRU state file."""
-        if self.agent_type == AgentTypeName.REGISTERED and self.registration_id is not None:
-            return f"{_REGISTERED_PREFIX}{self.registration_id}"
-        return self.agent_type.value
 
 
 def resolve_builtin_harness(name: str) -> HarnessSelection | None:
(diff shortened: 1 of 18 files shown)


tools/sculpt/sculpt/harness.py
```diff
@@ -0,0 +1,154 @@
...
+# AGENT_TYPE_LABELS so the CLI offers the same names the UI does.
+BUILTIN_HARNESS_LABELS: dict[AgentTypeName, str] = {
+    AgentTypeName.CLAUDE: "Claude",
+    AgentTypeName.PI: "Pi",
+    AgentTypeName.TERMINAL: "Terminal",
+}
+
+_STATE_DIR_ENV_VAR = "SCULPT_STATE_DIR"
+_DEFAULT_STATE_DIR = "~/.sculpt"
+_STATE_FILE_NAME = "cli_state.json"
```
saeedn: We shouldn't need any new state to support the MRU harness. The backend already tracks it, so the sculpt command should be able to just omit the harness flag and let the backend pick the MRU value.

And to clarify the requirement: it is not "preserve the MRU harness chosen by sculpt". It is "preserve the MRU harness chosen by the user within the Sculptor application".

PR #85, opened 2026-06-18
Remove the "Open in Modal" panel context-menu action

saeedn:
## Original bug

> "browser tab -> right click -> open in modal, then the browser tab crashes"

Reported via Sentry; tracked as SCU-1525. The right-click panel context menu
offered an "Open in Modal" action that opened a panel inside a Dialog. For the
Browser panel, opening it in the modal crashed the browser tab. Per the ticket,
the action provides no real value, so the fix is to **delete the feature, not
repair the crash.**

## Root cause

`PanelContextMenu` rendered an "Open in Modal" item, wired by `SidebarIcon` to
set `modalPanelIdAtom`. `PanelModal` (a `Dialog` host mounted by `DockingLayout`)
read that atom and rendered the panel's component a second time inside the modal.
Mounting the Browser panel's webview a second time inside the Dialog is what
crashed the browser tab. The feature had no other entry point and offered
nothing over the docked panel.

## Fix

Delete the entire code path rather than repair the crash:

- the "Open in Modal" item and `onOpenInModal` prop in `PanelContextMenu`
- the modal handler, atom wiring, and prop pass-through in `SidebarIcon`
- the `PanelModal` host component and its stylesheet
- the `modalPanelIdAtom` state atom
- the `<PanelModal />` mount in `DockingLayout`
- the obsolete modal / "Open in Modal" cases in `DockingLayout.test.tsx`

"Move to" and "Configure panels…" are unchanged.

## Verification

- `just format`, `just check` (ratchets, typecheck, lint, file-hygiene), and
  `just test-unit` (backend + frontend Vitest, including `DockingLayout.test.tsx`)
  all pass.
- Before/after of the panel context menu were captured via `/auto-qa-changes`
  and reviewed during the run: the menu goes from
  *(panel label · Move to · Open in Modal · Configure panels…)* to
  *(panel label · Move to · Configure panels…)*.

No dedicated test is added — proving a deleted feature's absence doesn't warrant
one.

(Sent by Claude)


sculptor/tests/integration/frontend/test_panel_context_menu.py
```diff
@@ -0,0 +1,48 @@
...
+
+from sculptor.constants import ElementIDs
+from sculptor.testing.elements.panel_zones import PlaywrightPanelZonesElement
+from sculptor.testing.playwright_utils import start_task_and_wait_for_ready
+from sculptor.testing.sculptor_instance import SculptorInstance
+from sculptor.testing.user_stories import user_story
+
+
+@user_story("to no longer see the crash-prone 'Open in Modal' panel action")
+def test_panel_context_menu_has_no_open_in_modal(sculptor_instance_: SculptorInstance) -> None:
```
saeedn: Delete this test, there's no need for an integration test to prove that a feature was deleted

saeedn: Done in 3adff86af0 — deleted `test_panel_context_menu.py` and reverted the context-menu POM accessors that only existed to support it. The component-level check in `DockingLayout.test.tsx` already covers the panel context menu, so the absence is still guarded without a dedicated e2e test.

(Sent by Claude)

PR #119, opened 2026-06-20
[SCU-1589] Add a workspace-widget SDK surface; Linear ticket shortcut in the banner

trisiak:
## What & why

The next step for the Linear example plugin: a workspace-level widget that sits beside the PR button and offers a Linear ticket shortcut, defaulted from the same logic as the panel but overridable by assigning a different ticket from the panel. The plugin is the vehicle for exploring **what the SDK needs** to support such a contribution — so the bulk of this change is a new, deliberately-clean SDK surface, with the Linear plugin as its first consumer.

## New SDK surface — `registerWorkspaceWidget`

- `WorkspaceWidgetDefinition { id; component; collapsePriority? }` + `registerWorkspaceWidget` on `PluginHostApi`, exported from the SDK barrel, backed by a new `pluginWorkspaceWidgetsAtom`.
- Mounted like a panel — per-plugin error boundary, `PluginContext`, and `WorkspacePluginContext` (so the workspace SDK hooks resolve) — and symmetric with `registerPanel`/`registerOverlay` (replace-by-id, instance-scoped undo).
- **Named placement-agnostically on purpose:** the same registration is what a future per-workspace vertical-tabs layout can render. `collapsePriority` is documented as a host hint (lower = hidden first); a host without a collapsing container may ignore it.

## Host rendering

`WorkspaceBanner` renders contributed widgets as **direct children** of the same flex row as `PrButton`, each in a `data-collapse-priority` wrapper gated by `hiddenPriorities`, so they participate in the existing `useProgressiveCollapse` scheme — "flexibly collapsible based on priority, same container as the PR button."

## Linear plugin (the reference consumer)

- A compact banner shortcut showing one ticket: the branch (primary) ticket by default, overridable. Clicking opens it via `openExternal`.
- **Shared state:** a dedicated per-workspace `usePluginSetting("shortcut:<id>")` both panel and widget read, so the ticket reference stays consistent across the two surfaces. The default path reuses the panel's `["primary", branch]` query key, so they dedupe to one fetch.
- The panel gains a bookmark control to assign/clear the shortcut and highlights the current one.

## Tests

- Host unit test extended: `registerWorkspaceWidget` registers into the atom (with `collapsePriority`) and is cleared on unload.
- Integration test extended (`test_bundled_linear_plugin.py`): with a mocked Linear GraphQL API (via `page.route`, no real credentials/network) and an API key set, a workspace's banner widget resolves and shows the branch ticket id.
- **Tracked follow-up:** assigning a *different* ticket from the panel (the bookmark) and asserting the widget follows the shared `shortcut` setting — needs a multi-ticket Linear mock, noted as a TODO in the test.

## Validation done locally

Host `tsc` clean; full `npm run build` succeeds and emits the plugin bundle; `pluginManager.test.tsx` 18/18 pass; eslint + prettier clean; the new integration test passes `py_compile` + `ruff` (lint/format). The heavy Playwright integration test was **not** run locally — relying on CI here, which is part of why this is a draft.

(Sent by Claude)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eh8ySwWVDeecNA4ibDk88X)_


trisiak:
### Widget verified live in the banner

Ran the new integration test against a real Sculptor instance + browser locally — **`1 passed`** — and captured the contributed widget rendering in the workspace banner. The `# SCU-1234 ●` chip is the new `registerWorkspaceWidget` contribution (Hash icon + ticket identifier + the issue's Linear state dot):

Linear workspace widget in the banner``

The PR button isn't beside it in this shot only because the test's fixture repo isn't a GitHub/GitLab origin (so the banner's PR button is hidden via `shouldShowTargetBranch`) — the widget renders independently, as designed. The Linear GraphQL API is mocked in-test via `page.route`, so no real credentials/network are involved.

The screenshot is committed as plugin documentation (referenced from `plugins/README.md` next to the widget's description), so it also renders in the README — happy to drop it if you'd rather it not live in the tree.

(Sent by Claude)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #147, opened 2026-06-23
Fix Browser panel dropping the first navigation typed before the webview attaches

saeedn:
## Problem

In the Electron Browser panel, the first navigation a user types is silently dropped if the workspace's `<webview>` hasn't attached yet — the panel stays stuck on its initial `about:blank`. This is easy to hit right after switching to a freshly-created workspace, whose webview is created on the tab-switch and may not have fired `did-attach` by the time the user types a URL and presses Enter.

It also makes `test_browser_panel.py` flaky in CI with the "about:blank" signature (most visibly `test_browser_panel_in_page_state_preserved_across_workspace_switch`), because attach can take well over 10s under xvfb + software rendering.

## Root cause

The user-typed toolbar path calls `loadURL` the instant Enter is pressed:

```ts
const navigate = (url) => {
  const webview = webviewRef.current;
  if (!webview) return;
  webview.loadURL(url).catch(() => {});
};
```

Before the guest `<webview>` fires `did-attach`, `loadURL` throws *"WebView must be attached to the DOM and the dom-ready event emitted"*, and the `.catch(() => {})` swallows it — so the navigation is lost.

The agent-command path already guards this exact race in `BrowserViewSlot` (it gates on `webContentsId` and replays the command once the webview attaches). The user-typed path never got the same treatment. That asymmetry is the bug.

## Fix

Make `navigate()` robust to the attach race, mirroring the proven agent-command gate:

- When the webview is not yet attached (`webContentsId === null`), stash the requested URL in a ref instead of calling `loadURL` and swallowing the rejection.
- The `did-attach` handler replays the stashed URL once the guest attaches, overriding the initial `about:blank` load.
- Last write wins, so the user's most recent typed URL is the one that loads.

One file changed (+23/−1), well-commented to explain the attach race.

## Testing

- `just format`, `just check` (typecheck, lint, ratchets, file hygiene), and `just test-unit` (backend, frontend, foundation, sculpt) all pass.
- The reproducer is the existing electron integration test `test_browser_panel.py::test_browser_panel_in_page_state_preserved_across_workspace_switch`, which runs via the offload electron CI job; this change removes its attach-timing flakiness.

(Sent by Claude)

### DIFF
--- a/sculptor/frontend/src/pages/workspace/panels/browser/useBrowserWebview.ts
+++ b/sculptor/frontend/src/pages/workspace/panels/browser/useBrowserWebview.ts
@@ -33,6 +33,12 @@ export const useBrowserWebview = (
   // slot. The ref holds the running value so partial updates can merge.
   const statusRef = useRef<BrowserViewStatus>({ ...INITIAL_STATUS, currentUrl: initialUrl });
 
+  // A navigation requested before the guest <webview> has fired did-attach.
+  // loadURL throws "WebView must be attached to the DOM and the dom-ready
+  // event emitted" until then, so navigate() stashes the URL here instead of
+  // losing it, and handleDidAttach replays it once the guest attaches.
+  const pendingNavigationUrlRef = useRef<string | null>(null);
+
   // Latest-ref pattern: the event-listener effect runs once with [], so it
   // captures the originally-passed callbacks. Reading via ref keeps it
   // pointing at the most recent props on every event.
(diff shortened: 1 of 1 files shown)


trisiak:
Heads-up: I'd independently found + fixed this exact bug in #125 — but your version is cleaner (gates on the existing `statusRef.current.webContentsId` rather than an extra mirror ref), so I dropped mine and rebased #125's *test-side* hardening on top of yours.

One data point that may be worth a tweak to the "removes the timing flakiness" note: in an offload repro (30 isolated copies, `retry_count=0`), your product fix alone takes `test_browser_panel_in_page_state_preserved_across_workspace_switch` to **30/30 still flaking**. The dropped-loadURL failures are gone, but the test then fails on two *separate* test-side races your change doesn't touch — `Script failed to execute` (the test bridge evaluating into a guest that's mid-load) and `address bar about:blank` (the helper binding a stale URL input after the workspace switch). With #125's retries stacked on top it drops to **2/30**.

Not a blocker for #147 — your fix is the right product-side change and I'd like it to land. Just flagging so nobody expects that specific test to go green from this PR alone.

(Sent by Claude)

PR #149, opened 2026-06-23
[SCU-1566] Clamp activeIndex on workspace delete + settle prompt-nav scroll

trisiak:
Two independent flakes from the alpha scroll / prompt-nav cluster (the top unaddressed flake cluster in the latest net).

## 1. Product fix — clamp `activeIndex` when deleting the active workspace
`optimisticDeleteWorkspaceAtom` removed the active tab and left `activeIndex` at the invalid sentinel (`-1`). The follow-up navigation only re-points it *synchronously* on the `navigateToAgent` path; the `navigateToWorkspace` / new-workspace-fallback paths defer it to an async effect that can be starved — so the **persisted** `sculptor-tabs` state keeps a `-1` activeIndex while tabs still exist. On the next reload, `rootLoader` reads `order[-1]` and **bounces to `/ws/new` instead of the surviving tab** (a real user-facing bug).

Fix: clamp to a surviving neighbor in the same synchronous `set`; the following navigation refines it. Updates the existing unit tests (the `-1` behavior was asserted) and adds delete-last / delete-only coverage.

## 2. Test fix — settle the scroll before the first ArrowUp in prompt-nav setup
`_setup_three_prompt_chat`'s dot-click issues an async `scrollToIndex`; `Escape` clears the highlight synchronously but doesn't wait for the scroll to land. The keydown hook reads live `scrollTop` against a 20px tolerance, so a first `ArrowUp` before the scroll settles is misread as "scroll current turn to top". Wait for the last prompt to pin to the viewport top before returning — fixes the whole `_wait_for_highlight_on` group (~5 tests).

## Validation (offload, isolated, `retry_count=0`)
| | flaked |
|---|---|
| prompt-nav (escape / arrow-up / enter) | **0/15 each** (was flaky) |
| clamp | **5/40** (was 11/40) |

The clamp product fix also clears the reload bug. **Deferred follow-ups:** a residual order-settle case in the clamp test (5/40), and the separate scroll-restore flake `test_first_message_visible_after_agent_switch` (a virtualizer re-measurement race — moderate-risk product change).

(Sent by Claude)

### DIFF
--- a/sculptor/frontend/src/common/state/atoms/workspaces.ts
+++ b/sculptor/frontend/src/common/state/atoms/workspaces.ts
@@ -623,8 +623,25 @@ export const optimisticDeleteWorkspaceAtom = atom(null, (get, set, workspaceId: 
   // arriving before the server confirms deletion doesn't treat it as
   // a "new" workspace and auto-open it as a tab. workspacesArrayAtom
   // filters out null atoms, so it won't appear in the UI.
-  // Remove from tab order so the tab disappears immediately
-  set(tabsAtom, applyClose(get(tabsAtom), workspaceId));
+  // Remove from tab order so the tab disappears immediately.
+  //
+  // When the deleted workspace was the active tab, applyClose leaves
+  // activeIndex at INVALID_ACTIVE_INDEX (-1). The navigation that follows
+  // usually re-points it at a surviving tab, but only synchronously on the
+  // navigateToAgent path; the navigateToWorkspace / new-workspace-fallback
+  // paths defer it to an async effect that can be starved, leaving the
+  // *persisted* tabs state at -1 while tabs still exist. On reload, rootLoader
+  // then reads order[-1] === undefined and bounces to /ws/new instead of
+  // restoring the surviving tab. Clamp to a valid neighbor synchronously so the
+  // persisted state is never invalid; the following navigation refines it.
+  const beforeClose = get(tabsAtom);
+  const removedIndex = beforeClose.order.findIndex((e) => e.tabId === workspaceId);
+  const afterClose = applyClose(beforeClose, workspaceId);
+  const nextTabs =
+    afterClose.activeIndex === INVALID_ACTIVE_INDEX && afterClose.order.length > 0
+      ? { ...afterClose, activeIndex: Math.min(removedIndex, afterClose.order.length - 1) }
+      : afterClose;
+  set(tabsAtom, nextTabs);
   // Track the deletion so components with their own workspace lists
   // (e.g. RecentWorkspaces) can filter it out without a page reload.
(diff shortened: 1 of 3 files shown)


sculptor/frontend/src/common/state/atoms/workspaces.ts
```diff
@@ -623,8 +623,25 @@ export const optimisticDeleteWorkspaceAtom = atom(null, (get, set, workspaceId:
...
+  // restoring the surviving tab. Clamp to a valid neighbor synchronously so the
+  // persisted state is never invalid; the following navigation refines it.
+  const beforeClose = get(tabsAtom);
+  const removedIndex = beforeClose.order.findIndex((e) => e.tabId === workspaceId);
+  const afterClose = applyClose(beforeClose, workspaceId);
+  const nextTabs =
+    afterClose.activeIndex === INVALID_ACTIVE_INDEX && afterClose.order.length > 0
+      ? { ...afterClose, activeIndex: Math.min(removedIndex, afterClose.order.length - 1) }
+      : afterClose;
+  set(tabsAtom, nextTabs);
```
trisiak: Should the transition of the active index be moved into `applyClose` ?

trisiak: Good call — done in 36f7bbaa149. Moved the transition into `applyClose` and reverted `optimisticDeleteWorkspaceAtom` to a plain `applyClose(...)` call.

It does belong there: leaving `activeIndex` at `-1` while tabs survive is an invalid state regardless of *which* close path produced it, and `applyClose` is the one place that computes the next tab state on removal. Centralizing means the invariant ("activeIndex is never the sentinel while a tab survives") now holds for all callers, not just optimistic delete.

I checked the blast radius across the other callers before moving it:
- **`updateWorkspacesAtom` (server-confirmed delete, e.g. from another client)** — now recovers the active pointer the same way the optimistic path does, instead of stranding at `-1`. Strict improvement; added a test for the active-tab-deleted case.
- **`closeWorkspaceTabAtom` (real tabs)** — unaffected; that path is reversible and uses `pendingClose`, not `applyClose`.
- **`openWorkspaceTabAtom` rollback** — the rolled-back tab was inserted with `setActive: false`, so it's never the active tab; no change.
- **`convertHomeTabToWorkspaceAtom` dedup** — only dedups a duplicate of the workspace being opened; the caller overrides `order` right after and a navigation sets the active workspace, so no behavior change in practice.

One intentional behavior nuance worth flagging: closing the active tab when **only pseudo-tabs survive** (e.g. just Home) now lands on that neighbor instead of bouncing to `/ws/new`. That seemed strictly better, but easy to special-case if you'd rather preserve the old `/ws/new` fallback there.

(Sent by Claude)

PR #156, opened 2026-06-24
Bound pi stdin writes so a stalled pi cannot wedge the control loop

DanverImbue:
## Summary

Fixes SCU-1594: `set_model` (and `/clear`) could hang indefinitely when the pi process stalls.

**Diagnosis correction.** The ticket framed this as an unbounded stdin write, but `RunningProcess.write_stdin` in this repo is non-blocking — it enqueues to a dedicated writer thread, and the pi agent only ever runs on it — so the write never wedges the control loop. (Thanks to the review that caught this.)

**Real cause.** A pi agent's control commands (`set_model`, `/clear`) and chat turns share one between-turns FIFO loop (`_process_message_queue`). A turn drains pi's stdout in `_consume_until_turn_end`, which has no deadline. If pi stays alive but emits nothing (a wedged turn), that loop never returns, so a queued `set_model` is never reached and the web layer's `await_request_outcome` waits on its outcome forever ("Waiting for outcome…").

## Fix

pi sends no heartbeat, so a silent turn is indistinguishable from a long but healthy quiet tool (e.g. a silent build). Killing the turn on silence would risk aborting legitimate work and desyncing Sculptor from a pi that is still running it. So the fix is **non-destructive**:

- `_consume_until_turn_end` tracks the time of pi's last stdout line.
- When the turn's pi has produced no output for `_TURN_STALL_TIMEOUT_SECONDS` (120 s, a single tunable constant) **and** a control command is queued behind it, `_fail_control_commands_starved_by_stall` resolves that command as a failed request with an actionable message ("The agent is busy or not responding. Stop the current turn and try again.").
- The turn and the pi process are **left untouched**. A genuinely wedged pi is recovered by the user pressing **Stop** (the existing abort + SIGTERM escalation); a chat turn queued alongside is put back, not dropped.

A `_control_command_enqueued` flag gates the queue scan so a turn with nothing waiting never pays for it. Healthy streaming turns reset the silence timer continuously, so a control command queued behind them still applies between turns exactly as before.

**Trade-off (intentional):** a model switch / context reset queued during a turn whose pi is genuinely silent for >120 s will fail (recoverable) rather than apply-after. Only a turn that emits nothing at all for the full window is affected, and the threshold is easy to tune.

The earlier stdin-write/lock bound from the first commit — which bounded a non-blocking call — is reverted in favor of this.

## Tests

- `test_stalled_turn_fails_a_starved_set_model_instead_of_hanging` — the headline path.
- `test_stalled_turn_fails_a_starved_clear_context` — the same guard covers context reset.
- `test_stall_failure_preserves_a_queued_chat_turn` — a chat turn queued alongside is not dropped.
- `test_normal_turn_leaves_a_queued_set_model_for_between_turns_processing` — a turn that ends normally does not trigger the guard.

## Verification

- `just format`, `just check` (ratchets / lint / typecheck) — green
- `just test-unit-backend` — green
- pi_agent suite (304 tests) — green

(Sent by Claude)

### DIFF
--- a/sculptor/sculptor/agents/pi_agent/agent_wrapper.py
+++ b/sculptor/sculptor/agents/pi_agent/agent_wrapper.py
@@ -262,6 +262,21 @@ _IDLE_WAIT_SECONDS: float = 1.0
 # get_state responses before giving up (see _fetch_models_into_state).
 _MODEL_FETCH_TIMEOUT_SECONDS: float = 10.0
 
+# How long pi may produce NO stdout during a turn before a control command
+# (model switch / context reset) queued behind that turn is failed rather than
+# left to wait on it. Control commands run strictly between turns on the single
+# `_process_message_queue` loop, so a turn that never ends starves them forever —
+# a queued model switch then never completes. pi emits no heartbeat, so total
+# silence is the only wedge signal available — and a long, genuinely-quiet tool
+# (a silent build) is indistinguishable from a wedge. The window is therefore
+# generous, and the response is non-destructive: only the starved command is
+# failed (the user retries, or presses Stop to recover a truly wedged pi via the
+# abort + SIGTERM escalation); the turn and the pi process are left untouched.
+_TURN_STALL_TIMEOUT_SECONDS: float = 120.0
+# Surfaced to the user when a control command is failed because the in-flight
+# turn's pi has gone silent past `_TURN_STALL_TIMEOUT_SECONDS`.
+_STALLED_CONTROL_COMMAND_MESSAGE: str = "The agent is busy or not responding. Stop the current turn and try again."
+
 # Transient-provider-error retry policy. A turn that ends with a known-transient
 # provider failure (overloaded / rate-limit / 5xx / timeout — see
 # `is_transient_provider_error`) is re-prompted up to this many times with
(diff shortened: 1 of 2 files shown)


DanverImbue:
Thanks @copilot — all four comments were correct, and they prompted a pivot of this PR.

You were right that `RunningProcess.write_stdin` is non-blocking (it enqueues to a dedicated writer thread), and the pi agent only ever runs on it, so bounding the write didn't bound a real blocking point and the tests stubbed a stall that can't occur in production.

The actual cause of the hang is the single between-turns loop: a turn drains pi's stdout in `_consume_until_turn_end`, which has no deadline, so a wedged-but-alive pi blocks the loop and a queued `set_model` is never reached. The push now replaces the stdin-write bound with a stall guard there: when the in-flight turn's pi has emitted nothing for `_TURN_STALL_TIMEOUT_SECONDS` and a control command is queued behind it, that command is failed with an actionable error. Because pi sends no heartbeat (a silent turn is indistinguishable from a quiet-but-healthy build), the response is deliberately non-destructive — the turn and pi process are untouched; a truly wedged pi is recovered by pressing Stop. See the updated PR description for details.

(Sent by Claude)

PR #161, opened 2026-06-24
SCU-1598: fetch Vault secrets via the public imbue-ai/use-vault-secrets action in CI

trisiak:
## What

Replace our vendored `scripts/export-vault-env` with the public [`imbue-ai/use-vault-secrets`](https://github.com/imbue-ai/use-vault-secrets) composite action for fetching Vault secrets in GitHub Actions ([SCU-1598](https://linear.app/imbue/issue/SCU-1598)).

The action authenticates with the same GitHub OIDC flow, talks to the Vault HTTP API directly with `curl` + `jq` (no Vault CLI, no download), writes secrets into `$GITHUB_ENV` (masked), and outputs the pinned secret versions for reproducibility.

> **Why a dedicated public action repo?** `imbue-ai/sculptor` is public, and a public repo's workflow cannot resolve an action that lives in a private repo. The action therefore ships from the public `imbue-ai/use-vault-secrets` repo (the Vault instance, its docs, and the Terraform that binds OIDC roles stay private in `imbue-ai/vault`).

## Changes

**Both workflows that read Vault secrets** are converted; each in-shell `eval "$(./scripts/export-vault-env ...)"` becomes a preceding `use-vault-secrets` step that injects the secrets into `$GITHUB_ENV` for the following work step.

- **`offload.yml`** (2 sites) — Modal credentials for the two integration-test jobs.
- **`build-desktop.yml`** (9 sites) — macOS signing certs + Apple notarization IDs; AWS creds; production-vs-dev frontend telemetry keys (gated on `env.CI_COMMIT_TAG`); S3-publish AWS creds; and the Slack webhook for the release / dev-release / build-failed notifications.

**Cleanups enabled by dropping the Vault CLI dependency** (the action uses only `curl` + `jq`):
- Removed all 5 "Install Vault CLI" steps (Linux apt, macOS brew, three release-job apt installs).
- Dropped now-dead packages: `unzip` (offload) and `gnupg` + `lsb-release` (build-desktop) — they only supported the Vault apt/zip install.
- Removed the redundant top-level `VAULT_ADDR` / `VAULT_NAMESPACE` env (the action sets them itself).
- **Deleted `scripts/export-vault-env`** — no references remain.

The old build step swallowed a missing-telemetry-key failure under `eval` then re-asserted the vars were non-empty; the action fails the step directly (`curl --fail`) when a key is missing, so that manual guard is gone.

## Verification

`offload.yml` runs on `pull_request`, so this PR exercises the action end-to-end. The `Set up job` (action resolution) and `Export Modal credentials` (OIDC → Vault auth → secret injection) steps both succeed on this PR's offload run. `build-desktop.yml` only runs on push-to-main / tags, so it's validated after merge.

(Sent by Claude)


DanverImbue (approved):
LGTM

PR #176, opened 2026-06-25
Make the pi/Claude model switcher provider-aware with cascading menus

DanverImbue:
## What

Rework the chat model switcher (shared by the pi and Claude harnesses) from a flat per-provider grouping into a single `DropdownMenu` with provider-aware states:

- **No providers** (pi, unconfigured): an "Authenticate a provider" prompt linking to pi settings, instead of silently falling back to the Claude list.
- **One provider**: a non-selectable provider header over its models (unchanged behavior).
- **Two or more providers**: a top-level entry per provider, each cascading into its own models.
- **Claude** (no backend catalog): the built-in model list, unchanged.

## Why

The switcher rendered every pi provider as a flat group header, and an unconfigured pi (no providers) was indistinguishable from a Claude task — both surfaced an empty model list and silently fell back to the built-in Claude models, with no nudge to authenticate.

## How

- Add a `sources_backend_models` harness method, surfaced as a computed field on the task view (base `false`, pi `true`) and read on the frontend through a narrow Jotai atom. This is the stable signal that tells "pi with no providers" apart from "Claude".
- Share the provider-grouping and Claude-list helpers between the two render paths.
- The generated API types and sculpt client pick up the new field; the sculpt CLI test fixtures are updated to include it.

## Testing

- Unit tests for all four render states (cascade sub-triggers, single-provider grouping, the authenticate prompt and its click handler, the Claude flat list, and radio selection) plus the backend harness method.
- `just format` / `just check` / frontend `just test-unit` green. Backend unit tests pass aside from the known-flaky `spawned_pty` terminal tests, which are unrelated to this change.

## Commits

1. The feature.
2. A standalone fix for Storybook's `~`-alias resolution — it failed to resolve `~/...ts` imports, which blocked every story.
3. A `ModelSelector` Storybook story covering the four states (used to visually verify the change).

(Sent by Claude)


DanverImbue:
image

PR #183, opened 2026-06-25
OpenHost: nginx /proxy front + Vite live-preview of the web UI

trisiak:
> **Stacked on #266** (`OpenHost enablement`). This PR targets the
> `maciek/openhost-enablement` branch and should merge **after** #266; its diff
> here is only the nginx `/proxy` front + Vite live-preview on top of that base.

## What

Fronts the OpenHost container with **nginx** so the Sculptor web UI can be
**previewed live (HMR) from any browser — including a phone — behind OpenHost's
owner SSO**, with nothing exposed publicly.

- nginx owns the OpenHost-proxied port **:5050**; the backend moves to loopback
  **:5051**. Routing: `/proxy/<port>/*` → `127.0.0.1:<port>` (dev servers, ports
  51000–59999); everything else → the backend.
- A preview is a Vite dev server served under `base=/proxy/<port>/`, dialing HMR
  back over `wss://…:443/proxy/<port>/`. `/api` + `/ws` stay same-origin absolute,
  so they reuse the one shared backend — **frontend-only** previews, no
  per-workspace backend.

## Changes

- `openhost-nginx.conf` *(new)* — the front + `/proxy/<port>/` reverse proxy
  (loopback-only upstream, 51000–59999 port band, session cookie stripped, WS upgrade).
- `openhost.Dockerfile` — install `nginx-light`; backend → `SCULPTOR_BIND_HOST=127.0.0.1`,
  `SCULPTOR_API_PORT=5051`.
- `openhost-run.sh` — launch the backend behind the nginx front.
- `openhost.toml` — document the nginx front / port split.
- `sculptor/frontend/vite.base.config.ts` — env-gated (`SCULPTOR_OPENHOST_PROXY`)
  base + wss HMR. No app code changes (same-origin API base; gated so normal
  dev/build is unaffected).
- `sculptor/frontend/launch-preview.sh` *(new)* — one-command preview launcher.
- `.claude/skills/openhost-frontend-preview/SKILL.md` *(new)* — OpenHost-only skill
  teaching agents to run it.
- `openhost-agent/` — the env skill + `AGENTS.md` gain the `/proxy` capability docs.

## Security

All behind OpenHost owner SSO (the only public ingress). `/proxy` adds
defense-in-depth: loopback-only upstreams (the OpenHost router is unreachable),
the 51000–59999 port band (can't target the backend itself), and the Sculptor
session cookie is stripped before reaching a preview. Full-stack (own-backend)
previews aren't supported on one origin — the session cookie would collide.

(Sent by Claude)

### DIFF
--- a/.claude/skills/openhost-frontend-preview/SKILL.md
+++ b/.claude/skills/openhost-frontend-preview/SKILL.md
@@ -1,0 +1,85 @@
+---
+name: openhost-frontend-preview
+description: |
+  Live, hot-reloading preview of the Sculptor WEB frontend from a phone or any
+  browser, served through the OpenHost nginx /proxy front. ONLY relevant when
+  running INSIDE the openhost-deployed Sculptor (the container fronted by nginx on
+  :5050, reachable at https://sculptor.<zone>/) — NOT for local or Electron dev.
+  Frontend-only: it reuses the single shared backend, so it previews UI/frontend
+  changes, not a per-workspace backend. Invoke when iterating on the web UI and
+  wanting to see it live in a browser through the OpenHost SSO front.
+user-invocable: true
+---
+
+# OpenHost frontend live preview (`/proxy/<port>/`)
+
+Run a Vite dev server (HMR) for the Sculptor web UI and reach it from any
+logged-in browser — including a phone — through the OpenHost nginx front, entirely
+behind OpenHost's owner SSO. Nothing is exposed publicly.
+
+## When this applies — OpenHost only
+
+Use this ONLY inside the openhost-deployed Sculptor. Quick checks:
+- `test -f /app/openhost-nginx.conf` (the nginx front config), and
+- PID 1 is nginx: `ps -p 1 -o args=` -> `nginx ...` (backend is on loopback :5051).
+
+If those are NOT true (local dev, Electron, a plain backend on :5050), this skill
+does not apply — use normal `pnpm run dev` / `just frontend` there instead.
+
+## How it works (read first)
+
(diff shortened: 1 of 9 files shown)


trisiak:
Review follow-up — addressed in `dc4bd96` (plus earlier cleanup). The threads above are resolved; summarizing here because GitHub blocks per-thread replies while a pending review draft is open on the PR:

- **nginx header — branch name:** removed; the header now describes the config's purpose (nginx front + `/proxy` live preview) without naming a branch or individual.
- **nginx header — external "workspace repo" note:** the full-stack-preview limitation is now stated inline; the external pointer is gone, so the rationale is self-contained.
- **`docs/openhost-dev-preview.md` (branch-specific / external pointers):** the doc was removed entirely. Usage now lives in the `openhost-frontend-preview` skill, and the design rationale is inline in `openhost-nginx.conf` / the shell-script headers (closer to the code, less drift).
- **`openhost-run.sh` — process supervision:** intentionally keeping a single `exec nginx` (PID 1 = nginx). The manifest declares a `/api/v1/health` check, so a dead backend fails the probe and the platform restarts the app; happy to add explicit supervision if that ever proves insufficient.
- **`vite.base.config.ts` — HMR / env naming:** `base` is now passed natively (`vite --base=/proxy/<port>/`); only the no-CLI-flag HMR-over-TLS bits remain in config, gated on the explicitly-named `SCULPTOR_OPENHOST_PROXY`. Also fixed the pytest gate so it only *disables* HMR rather than clobbering that override.

PR #200, opened 2026-06-26
Add homepage plugin slot with a persisted view switcher (SCU-1610)

trisiak:
## What

Adds a **homepage plugin slot**: plugins can contribute a full-page "home view" that the homepage offers as an alternative to the built-in recent-workspaces list.

When at least one home view is registered, the homepage shows a switcher between the built-in **Recent workspaces** view and the plugin views. The selected view takes over the homepage body. The choice is remembered, and falls back to the built-in view if the selected plugin is no longer registered. With no plugin home views, the switcher is hidden and the homepage is unchanged.

## How

- **New contribution point.** `HomeViewDefinition` (`id`, `title`, optional `icon`, `component`) + `registerHomeView` on the plugin host API, a `pluginHomeViewsAtom` registry, and the manager wiring. Home views are app-global like overlays — wrapped in the per-plugin error boundary and `PluginContext`, with **no** `WorkspacePluginContext` (the homepage isn't workspace-scoped).
- **Built-in view extracted.** The existing recent-workspaces homepage moves into `RecentWorkspacesHomeView`. `HomePage` now hosts the switcher and renders the selected view.
- **Persistence.** The selection is stored in `localStorage` (consistent with how plugin sources are persisted), behind a pure `resolveEffectiveHomeViewId` fallback resolver with unit tests.
- **Switcher.** A Radix `SegmentedControl`, shown only when there's more than one view to choose between.

The example/demo home-view plugin shown during development is intentionally **not** included here — it will land alongside the Linear plugin work.

## Testing

- Unit tests for the fallback rule and the derived option/effective-view atoms (`homeViews.test.ts`).
- `just check` (typecheck + lint + ratchets + hygiene) and the frontend unit suite pass.
- Manually QA'd end-to-end via the live-browser harness with a throwaway home-view plugin: baseline unchanged with no plugin; switcher appears once a home view registers; selected view renders and switches both ways; selection persists across in-session navigation.

(Sent by Claude)

### DIFF
--- a/agent_docs/scu-1610-homepage-plugin-slot/mocks.html
+++ b/agent_docs/scu-1610-homepage-plugin-slot/mocks.html
@@ -1,0 +1,411 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>SCU-1610 — Homepage plugin slot mocks</title>
+    <style>
+      :root {
+        --gray-1: #111113;
+        --gray-2: #18191b;
+        --gray-3: #212225;
+        --gray-4: #292a2e;
+        --gray-5: #303136;
+        --gray-6: #3a3b40;
+        --gray-border: #2b2c30;
+        --text-hi: #ededef;
+        --text-mid: #b0b0b8;
+        --text-lo: #7c7d85;
+        --accent: #5b9dff;
+        --accent-dim: #2c3a52;
+        --green: #46c46e;
+        --amber: #e0a23b;
+        --red: #e5564b;
+        --blue: #5b9dff;
+        --radius: 10px;
+        font-synthesis: none;
+      }
+      * { box-sizing: border-box; }
+      html, body { margin: 0; height: 100%; }
+      body {
+        background: var(--gray-1);
+        color: var(--text-hi);
+        font-family: ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto, sans-serif;
+        font-size: 14px;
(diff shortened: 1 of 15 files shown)


trisiak:
Addressed the Copilot review in edc9f0703bb.

Both comments were the same reserved-id edge case: a plugin registering a home view with the built-in id `recent-workspaces` would create duplicate option ids (duplicate React keys + duplicate `SegmentedControl` values). Fixed on two layers:

- `registerHomeView` now rejects that id up front with a `console.warn`, so the plugin author gets a reason the view didn't appear (rather than it silently vanishing).
- `homeViewOptionsAtom` filters it out as a defensive second guard.

Added a unit test pinning that the option ids stay unique and the built-in view keeps the reserved id.

(Sent by Claude)

trisiak:
(going in with https://github.com/imbue-ai/sculptor/pull/218)

trisiak:
Closing — this PR's content is already in `main`. It landed via #218 (whose branch carried the whole stack), which merged **patch-identical copies** of these commits rather than these exact SHAs, so GitHub couldn't auto-mark this PR merged.

Verified: a simulated merge of this branch into `main` yields a byte-identical tree (a no-op), and every commit here is patch-equivalent to one already in `main`:

| This branch | Already in `main` (via #218) |
| --- | --- |
| `93e85fc` Add homepage plugin slot with a persisted view switcher (SCU-1610) | `23a6fac` |
| `edc9f07` Reject plugin home views that claim the reserved built-in id | `21f7797` |
| `a3398022` Extend the plugin SDK (and home-view switcher) for the board (from #232) | `a4d5613` |

Nothing left to merge; closing and deleting the branch.

(Sent by Claude)

PR #206, opened 2026-06-27
[SCU-1609] Enable frontend plugins by default with Linear bundled

trisiak:
## What

Enable the frontend plugin system by default, with only the bundled Linear plugin enabled out of the box.

- Flip the `enable_frontend_plugins` user-config default `false` → `true`, so the plugin system loads on a fresh install.
- Only the bundled **Linear** plugin is enabled by default; the `sculpty` and `pomodoro` example plugins ship `disabledByDefault` and can be turned on per plugin.
- Move the **global enable/disable toggle** for the plugin system out of **Experimental** into the **Plugins** settings section. The Plugins section is now always visible (it hosts the toggle, so it must stay reachable to turn the system back on); turning the toggle off hides the plugin-management UI while the section and toggle stay in place.
- The bundled Linear **panel** ships **opt-in** (`defaultEnabled: false`) — registered but enabled from Settings → Panels (like the built-in Browser panel), while its always-on **banner widget** is the default surface. An on-by-default plugin shouldn't claim a slot in everyone's panel layout uninvited.
- A failed plugin load now exposes a **Retry** control on its row (a one-shot manual reload, not a retry loop), so a plugin that fails to load (e.g. on a cold dev-server start) can be recovered in place without an app reload.
- Fix a latent layout race the plugins-on default exposed: a plugin registering its panel during the layout bootstrap could leave every docked zone hidden (collapsing the file browser / terminal / etc. to just the center panel). The bootstrap now seeds each layout piece independently. Includes a regression unit test.

## Why

The plugin system is ready to ship on by default, with Linear as the flagship bundled plugin. A global on/off toggle co-located with the rest of the plugin controls keeps a clean kill switch without an Experimental opt-in, and shipping the Linear panel opt-in keeps the default layout uncluttered.

## Testing

The integration suite now runs plugins-on, matching production. Because the Linear panel is opt-in it no longer perturbs panel-layout tests; the tests that drive the panel enable it explicitly. Verified locally: the plugin loader/retry, bundled-Linear, panel-zones, panel-open-state, plugin-runtime, and Plugins-visibility integration tests pass; `format` / `check` / unit-tests green. (One unrelated offload-runner flake — `test_file_browser`, "Repo Path Not Found" — passes locally and is being investigated separately.)

## Docs

`docs/specs` updated: a new `SPEC.md` §7.13 describes the plugin system as generally available, with the global toggle in the Plugins section and the Linear panel as opt-in; `REQ-FUNC-016` / `REQ-SEC-004` record the default-on flag; scenarios `SET-040` (global toggle) and `SET-041` (retry a failed plugin load) were added.

Linear: SCU-1609

(Sent by Claude)

### DIFF
--- a/sculptor/tests/integration/frontend/test_plugin_loader.py
+++ b/sculptor/tests/integration/frontend/test_plugin_loader.py
@@ -412,18 +412,19 @@ def test_plugin_loader_in_electron(sculptor_instance_: SculptorInstance) -> None
     differ). That packaged ``file://`` case remains a separate, unsolved scenario.
 
     The factory fixture can't launch a non-packaged Electron, so Electron
-    coverage rides the shared instance. That instance ships with the flag off, so
-    we flip it on through the Experimental settings (live, no reload) before
-    driving the Plugins section, and flip it back off afterward so later
-    Electron tests start from a clean flag. (The error rows the run leaves behind
-    are renderer-local and harmless -- no other Electron test reads plugin
-    sources -- so we don't bother removing them.)
+    coverage rides the shared instance. The broad UI suite pins the plugin
+    system off (see ``_make_test_user_config``), so we flip it on through the
+    Plugins section's master switch (live, no reload) before driving the section,
+    and flip it back off afterward so later Electron tests start from a clean
+    flag. (The error rows the run leaves behind are renderer-local and harmless
+    -- no other Electron test reads plugin sources -- so we don't bother removing
+    them.)
     """
     settings_page = navigate_to_settings_page(page=sculptor_instance_.page)
-    settings_page.click_on_experimental().set_frontend_plugins(enabled=True)
+    plugins = settings_page.click_on_plugins()
+    plugins.set_frontend_plugins(enabled=True)
     try:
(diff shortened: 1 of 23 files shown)


sculptor/tests/integration/frontend/test_plugin_loader.py
```diff
@@ -412,23 +441,24 @@ def test_plugin_loader_in_electron(sculptor_instance_: SculptorInstance) -> None
...
+    Plugins section's master switch (live, no reload) before driving the section,
+    and flip it back off afterward so later Electron tests start from a clean
+    flag. (The error rows the run leaves behind are renderer-local and harmless
+    -- no other Electron test reads plugin sources -- so we don't bother removing
+    them.)
     """
     settings_page = navigate_to_settings_page(page=sculptor_instance_.page)
-    settings_page.click_on_experimental().set_frontend_plugins(enabled=True)
+    plugins = settings_page.click_on_plugins()
+    plugins.set_frontend_plugins(enabled=True)
```
trisiak: We don't need this anymore as the default is getting swapped around.

PR #208, opened 2026-06-27
Add `sculpt debug heap`: an opt-in backend memory-profiling endpoint (SCU-1625)

trisiak:
## What

Adds a development-only heap-introspection endpoint and `sculpt` subcommand for diagnosing `sculptor_backend` RSS growth, alongside the existing `/api/v1/debug/threads` and `/api/v1/trace/*` tooling.

- **`GET /api/v1/debug/heap`** (auth-gated; `include_in_schema=False` so it does not churn generated frontend types):
  - One-shot **gc-based object census** — top types by live count and aggregate shallow size, plus `gc` stats. `limit` bounds how many objects are sized so the GIL pause stays small on a multi-GB heap.
  - **`?collect=true`** forces a full `gc.collect()` and reports RSS before vs. after, which distinguishes accumulating cyclic garbage from genuine live retention.
  - **`?start_trace=N` / `?stop_trace=true`** toggle `tracemalloc` at runtime. The packaged backend is a PyInstaller bundle that ignores `PYTHONTRACEMALLOC`, so launch-time enabling isn't an option — this lets you turn allocation tracking on, reproduce the growth, then read back the top allocation sites both **by line** and **by full caller traceback**.
- **`sculpt debug heap [--collect] [--start-trace [--trace-frames N]] [--stop-trace] [--top N] [--limit N] [-o FILE]`** — thin CLI over the endpoint with a raised request timeout for the census pause.

## Why

This is the tooling used to find and confirm the recent backend RSS-growth regression. It is being split out as a standalone, mergeable change so the memory profiler can land independently of the fix it helped find.

## Safety

Inert until invoked: the census runs only on request and `tracemalloc` stays off unless explicitly started, so there is no always-on cost. Auth-gated and hidden from the public API schema. Development-only diagnostic.

## Test

`sculptor/sculptor/web/debug_heap_endpoint_test.py` — census returns, `collect` reports an RSS delta, `top`/`limit` are honored, and runtime tracemalloc start/stop produces the "by full traceback" section. 4 passed locally; `just check` / pre-commit green.

(Sent by Claude)

### DIFF
--- a/sculptor/sculptor/web/app.py
+++ b/sculptor/sculptor/web/app.py
@@ -4589,6 +4589,13 @@ def get_debug_heap(
     not atomic ``str``/``bytes``; their payload bytes show up in RSS and via
     tracemalloc, not the shallow-size column. Requires the session token.
     Development only."""
+    # Clamp to non-negative: a negative `top` would slice almost the whole list
+    # (Python's [:n] semantics) and dump a runaway report, and a negative `limit`
+    # would quietly disable the census cap.
+    top = max(0, top)
+    limit = max(0, limit)
+    start_trace = max(0, start_trace)
+
     process = psutil.Process()
     started_at = time.monotonic()
 
@@ -4625,7 +4632,9 @@ def get_debug_heap(
         counts[name] = counts.get(name, 0) + 1
         try:
             sizes[name] = sizes.get(name, 0) + getsizeof(obj)
-        except Exception:
+        except TypeError:
+            # getsizeof raises TypeError when an object's __sizeof__ returns a
+            # non-int; skip that one object rather than aborting the census.
             pass
     # Drop our strong reference to the full object list before formatting.
     del objects
--- a/sculptor/sculptor/web/app.py
+++ b/sculptor/sculptor/web/app.py
@@ -2,6 +2,7 @@ import asyncio
 import base64
 import contextlib
 import datetime
+import gc
 import json
 import logging
 import mimetypes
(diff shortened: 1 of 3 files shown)


trisiak:
Thanks for the review — addressed in 0953b00:

- **Negative `top`/`limit`/`start_trace`** — clamped to non-negative in the endpoint, so a negative `top` can't slice almost the whole type list into a runaway report and a negative `limit` can't disable the census cap. (Skipped an arbitrary max: the route is behind the session-token middleware, so the only caller is the local dev who can already do anything; the real fix here is the footgun, not a DoS surface.)
- **`except Exception` in the census** — narrowed to `except TypeError`, the documented way `sys.getsizeof` fails (an object whose `__sizeof__` returns a non-int), so a genuine bug surfaces instead of being swallowed per object.
- **`--trace-frames` with `--start-trace`** — the CLI now rejects `--trace-frames < 1` instead of sending `start_trace=0` and silently not starting tracemalloc.

Two I'm leaving as-is:

- **Auth-gating test** — auth is enforced globally by `SessionTokenMiddleware`, not per route, so a test here would exercise the middleware rather than this endpoint; the sibling `/api/v1/debug/threads` follows the same pattern with no per-route auth test.
- **Proof-of-work fenced test output** — that checklist category targets bug-fix MR bodies from an autonomous workflow; this is a tooling PR and the `(Sent by Claude)` line is the repo's transparency sign-off, not a workflow marker.

Added a regression test (`test_debug_heap_clamps_negative_params`) for the clamp.

(Sent by Claude)

PR #212, opened 2026-06-29
sculpt plugin commands: load / reload / inspect frontend plugins (SCU-1518)

trisiak:
**Draft — live-verified first pass, still under review.** Lets an agent develop a frontend plugin and load / reload / inspect it in the live Sculptor UI via `sculpt plugin`. Builds on the SCU-1517 drop-in plugin system.

Design doc: `docs/development/scu-1518-sculpt-plugin-commands.md`.

### What's here

**Backend bridge** — the `allow_agent_plugin_loading` switch (default off, sole consent gate for write ops); `PluginCommandUiAction` fanned over the existing WebSocket; a correlation bus so `POST .../plugins/command` broadcasts and gathers **per-renderer** replies; an install endpoint writing packaged plugins to `dev/<ws>/<id>/` (or top-level for `--persist`) with path-traversal + base64 guards; a remove endpoint. Read-only `inspect`/`list` ungated; write ops gated.

**Renderer** — handles the command in `useUnifiedStream`, runs it via `PluginManager`, replies with a self-reported identity `{rendererId, environment, origin}` and a **redacted** snapshot (status, origin, registrations, config key **names only**). Settings: an "Agent plugin loading" toggle + a "dev" badge on dev-origin rows.

**CLI** — `sculpt plugin load <path|url> [--persist] / reload / unload / remove / list / inspect / dir`; path loads are packaged + uploaded then loaded from the served manifest URL; multi-renderer results prefer the Electron window, `--all` / `--json` available.

**Dev-server fix** — proxy `/plugins/local` to the backend in the Vite dev server. Without it, *no* backend-served plugin (SCU-1517 drop-ins or SCU-1518 dev-loads) can load in `vite dev` mode — the loader's manifest fetch hit Vite's SPA fallback. Surfaced during live verification.

### Verification — live-verified end to end ✅
Ran the full round-trip against the manual-test harness (headless browser = a real connected renderer, both flags on):
- `sculpt plugin list` → real per-renderer reply with self-identity; builtin plugins reported `loaded`, with registration attribution (`linear-issue` → `panels:["Linear"], hasSettings:true`).
- `sculpt plugin load <dir>` → packaged → installed → broadcast → renderer fetched the served dev bundle → `activate()` ran → a banner rendered in the live DOM; `inspect` shows `loaded`, `origin=dev`.
- `reload` re-activates; `unload` runs the plugin's disposer (banner removed); `remove` deletes the dev files; the "dev" badge shows in Settings → Plugins.
- `just format` / `just check` / `just test-unit` all green; backend unit tests cover the bus, dev/persist placement, gating, id validation, and path-traversal.

### Deferred / follow-ups
- A scripted integration test mirroring `test_plugin_loader.py` (the live check above was manual via the harness).
- Browser mode beyond Electron-first; multi-window beyond the preferred-pick; host-side `sculpt`.
- Dev-source persistence across app restart (URL sources persist; served dev paths are re-pushed by the agent).
- Overlay→plugin attribution in `inspect`.

Linear: SCU-1518

(Sent by Claude)


trisiak:
Addressed all three Copilot review comments in bbcf0d893f5:

- **No-window delay** (`app.py` command endpoint): return immediately when `subscriber_count() == 0` instead of blocking for the full timeout.
- **Result correlation-id** (`app.py` result endpoint): reject a reply whose body `correlation_id` disagrees with the path (400), so a mismatched result can't reach the wrong waiting command.
- **Manifest-phase pluginId** (`pluginManager.tsx`): a manifest fetch/parse failure's synthetic `manifest.id` is the URL, not a real id, so it's no longer retained — the error state falls back to the source key (and the misleading comment is fixed). Only validate/import/activate failures, where the manifest parsed, keep the real id.

Added unit tests for the correlation-id mismatch and the manifest-phase id fallback; `just check` + unit tests green.

(Sent by Claude)

PR #225, opened 2026-06-30
Add living pi-integration reference under agent_docs/pi-integration (SCU-1637)

DanverImbue:
## What & why

`pi` is a third-party, version-pinned, immutable harness, but it had no
consolidated living reference — so each integration cycle re-derived the same
RPC / CLI / auth / skills facts from the code. This stands up
`agent_docs/pi-integration/` as that reference (SCU-1637).

The folder **summarizes and links** the de-facto live reference (the `harness.py`
and `agent_wrapper.py` module docstrings) rather than duplicating it, and **cites
a live source file for every non-obvious claim** so the doc has a fixed point to
be re-checked against as the code moves.

## Contents

- **README.md** — invariants: pi-core immutability, the pinned-version + pinned-extension model, the current pin, a capability snapshot, and how it relates to `agent_docs/pi-auth/` (and the not-yet-merged `pi-capabilities` work).
- **rpc.md** — the model-selection RPC trio (`get_available_models` / `get_state` / `set_model`); why model identity is `(provider, modelId)` fetched at runtime, not an internal enum; why no `--provider` / `--api-key`.
- **cli-flags.md** — the full launch argv; `--session-id` (not `--session`) and its adopt-verbatim / create-if-missing / corrupt-tail-tolerant resume behavior.
- **auth.md** — `~/.pi/agent/auth.json` as the shared source of truth; catalog gating on credential *presence* (not validity); the auth.json-vs-env precedence distinction; per-provider `/logout`; `getAgentDir` (`PI_CODING_AGENT_DIR`) resolution.
- **skills.md** — how Sculptor skills/commands map onto pi's `--skill`; un-namespaced plugin skills; loose `.claude/commands` wrapping; the no-description-doesn't-load rule; the full skill-source list.

## Notes for reviewers

- **Verified against live code, not the ticket.** Two facts in the source material were stale and are corrected here: the live pin is **`0.80.2`** (some in-code prose still says `0.78.0`), and the "auth.json > env" precedence is **pi's own** runtime resolution — Sculptor itself computes a *union* of the two sources for catalog gating. The doc states both layers distinctly.
- **One small functional addition.** Besides the docs, this adds a thin `/pi` skill (`.claude/skills/pi/SKILL.md`) that points at the reference so it's discoverable from the slash picker. It's a new slash command (handed to pi and Claude as a skill), so this PR is "docs-only" with that one caveat. No `sculptor/` code changes; no behavior change to pi-integration code itself.
- **Citations are line-pinned and will drift.** That's inherent to the approach and called out in the doc's own "verify before you trust" disclaimer plus a "keep this living" section.

## Verification

`just format`, `just check`, and `just test-unit` all run green (the gates need `node` on PATH via nvm and the worktree's frontend deps installed). `/crispy-comments` (no code comments in the diff) and `/code-review-checklist` (clean) were run before opening this PR.

(Sent by Claude)

### DIFF
--- a/.claude/skills/pi/SKILL.md
+++ b/.claude/skills/pi/SKILL.md
@@ -1,0 +1,37 @@
+---
+name: pi
+description: |
+  Point at the living pi-integration reference (agent_docs/pi-integration/).
+  Use at the start of any task that touches Sculptor's pi harness integration —
+  RPC model selection, CLI launch flags, provider auth, or skills mapping — so
+  you read the consolidated facts instead of re-deriving them from the code.
+---
+
+# pi Integration Reference
+
+`pi` is a third-party, version-pinned, **immutable** coding-agent harness that
+Sculptor drives as a subprocess. Sculptor never modifies pi-core; it extends pi
+only through a pinned extension set plus Sculptor-side wiring.
+
+Before changing or reasoning about any pi-integration surface, read the living
+reference under [`agent_docs/pi-integration/`](../../../agent_docs/pi-integration/README.md):
+
+- **[README.md](../../../agent_docs/pi-integration/README.md)** — invariants:
+  pi-core immutability, the pinned-version + pinned-extension model, the current
+  pin, and where the de-facto live reference lives (the `harness.py` /
+  `agent_wrapper.py` module docstrings).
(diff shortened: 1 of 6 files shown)


DanverImbue:
No need for this, we will document pi differently.

PR #234, opened 2026-07-01
Integration test: sculpt plugin load round-trip in a live renderer (SCU-1652)

trisiak:
**Stacked on #212** (base = `maciek/scu-1518-sculpt-for-plugins`). Draft until #212 merges.

Adds the first e2e coverage of the plugin **command bridge** against a real renderer — the full path that was previously only unit-tested (endpoints, PluginManager) and manually verified via the auto-qa harness.

`sculptor/tests/integration/frontend/test_plugin_command_roundtrip.py`:
- Browser-mode factory instance, seed `allow_agent_plugin_loading=True` via the config populator.
- Wait until the renderer's stream is connected (poll a read-only `list` command until it replies).
- Install a packaged plugin into the dev tree, then load it by its served manifest URL through `POST .../plugins/command`.
- Assert the endpoint returns the renderer's reply — a real `rendererId`, `ok: true`, and the plugin `loaded` from the `dev` origin — and that the renderer's own Settings → Plugins list shows it loaded (`data-status="loaded"`).

**Single window, browser mode only.** A live renderer here is one Playwright page whose `useUnifiedStream` opened a stream WS. Multi-renderer broadcast fan-out (two windows, distinct `rendererId`s) is intentionally deferred — it carries a harness unknown (a second BrowserContext) that shouldn't hold up the single-window coverage.

Runs green: `1 passed in 7.12s` via `just test-integration`.

Linear: SCU-1652

(Sent by Claude)

### DIFF
--- a/sculptor/tests/integration/frontend/test_plugin_command_roundtrip.py
+++ b/sculptor/tests/integration/frontend/test_plugin_command_roundtrip.py
@@ -17,6 +17,8 @@ import base64
 import json
 from pathlib import Path
 
+from playwright.sync_api import Error as PlaywrightError
+
 from sculptor.primitives.ids import WorkspaceID
 from sculptor.services.user_config.user_config import load_config
 from sculptor.services.user_config.user_config import save_config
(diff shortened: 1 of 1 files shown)


trisiak:
Ran the code-review checklist and Copilot's review; addressed both in 8ce976d101e:

- **Transient-error tolerance** (Copilot): `_wait_for_renderer` now retries on a Playwright request error / as-yet-empty reply while the page is still booting, instead of failing on a startup hiccup.
- **Request timeout** (Copilot + self): dropped the explicit `timeout=20_000` on the API POST so it uses the harness's 30s default — avoids an avoidable flake under CI load.
- **Empty-plugins guard** (Copilot + self): assert the reply carries a plugin before indexing, for a clearer failure message.
- **Magic numbers** (self): extracted the readiness-poll `40` / `500` into named constants.

Re-ran green: `1 passed in 8.45s`.

(Sent by Claude)

PR #245, opened 2026-07-01
PR Gate: fail open on transient API errors

trisiak:
## What

The contributor gate (`.github/workflows/pr-gate.yml`) now **fails open**: it closes a PR only when it can *positively* confirm the author is an uncleared external contributor. Any ambiguous failure leaves the PR untouched.

## Why

The gate auto-closed a maintainer's PR (#244) after a single flaky API call. `getPermission()` wrapped the collaborator-permission lookup in a bare `catch { return null }`, so a transient failure (network blip, 5xx, rate limit) was indistinguishable from "this user is not a maintainer." With github-script's default `retries: 0`, one bad call was enough to fall through to auto-close — and the run in question happened during visible network instability (the action tarball itself failed to download and had to back off). The `APPROVED_CONTRIBUTORS` read had the same flaw, silently degrading any read error into an empty list.

Closing a PR is disruptive and reads as a rejection, so the gate should never do it on incomplete information.

## Changes

- **`retries: 3`** on the github-script step, so ordinary transient 5xx/network errors never reach the catch.
- **`getPermission()`** returns `null` only on a definitive `404` (genuinely not a collaborator) and re-throws everything else.
- **Both call sites** (the permission lookup and the `APPROVED_CONTRIBUTORS` read) bail without touching the PR on any non-404 error, logging the status for the run log.
- **Header comment** documents the fail-open contract so it isn't reverted later.

## Scope

Single-file change to `.github/workflows/pr-gate.yml`. Verified with `actionlint` (clean), YAML parse, and an async-wrapped `node --check` of the embedded script; `just format` and `just check` pass.

(Sent by Claude)

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -9,6 +9,13 @@ name: PR Gate
 # Uses pull_request_target so it can comment/close on PRs from forks. It does NOT
 # check out or run any PR code, so untrusted code never executes here. Keep it
 # that way — see README.md in this directory (rule 1) before adding a checkout.
+#
+# Fails OPEN: closing a PR is disruptive and looks like a rejection, so the gate
+# only closes when it can POSITIVELY confirm the author is an uncleared external
+# contributor. If an API call fails for any reason other than a definitive 404
+# (transient 5xx, network blip, rate limit), the job leaves the PR untouched
+# rather than risk closing a maintainer's PR. github-script retries handle the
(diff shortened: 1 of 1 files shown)


danverbraganza (approved):
LGTM!

DanverImbue:
Noticed a flake when this tried to merge. I've readded it again, and dispatched an agent to work on the flake.

PR #254, opened 2026-07-02
Harden claude stream-json parser against unexpected message shapes

trisiak:
## Original bug

> **Harden claude stream-json parser against unexpected message shapes (turn-crash safety)** ([SCU-1667](https://linear.app/imbue/issue/SCU-1667/harden-claude-stream-json-parser-against-unexpected-message-shapes))
>
> A comparison of Sculptor's `claude` CLI wrapper against the official `claude-agent-sdk-python` found that Sculptor's stream-json parser can crash an entire agent turn on messages that are **valid JSON but have an unexpected shape**. The output loop only catches `json.JSONDecodeError` around parsing (`output_processor.py:566`); any other exception (`TypeError`, `KeyError`, `IndexError`) propagates out of the worker thread and kills the turn.
>
> Three concrete failure shapes were verified in current code:
> 1. `message.content` as a **string** on assistant messages → `TypeError` (`claude_state.py:432-433`).
> 2. **User message with an empty content list** → `IndexError` (`claude_state.py:459`).
> 3. `tool_result` **block without a** `content` **key** → `KeyError` (`claude_state.py:477`).
>
> Suggested fix (two layers): Layer 1 handles the known shapes in the parser; Layer 2 widens the containment `except` so the next unknown shape degrades to a visible warning instead of killing the turn.

**Test-kind note:** these are unit tests, not the default end-to-end tests. Per `.sculptor/testing.md`, e2e is required unless the bug "only manifests under conditions Playwright cannot produce" / the code path is internal-only. Both apply here: the buggy path is the internal stream-json parser consuming bytes from the Claude CLI subprocess's stdout, and these malformed shapes cannot be driven through the UI — real Claude does not emit them on demand (they are rare real-world edge cases; the official SDK only just fixed one on 2026-06-25), and `fake_claude` is a UI standin, not a faithful source of malformed CLI protocol bytes. The existing parser tests are all unit tests, so that is the natural home, and the ticket prescribes it.

## Hypotheses considered

1. **Assistant message with string `content` crashes the turn** — REPRODUCED (`TypeError` at `claude_state.py:433`).
2. **User message with an empty content list crashes the turn** — REPRODUCED (`IndexError` at `claude_state.py:459`).
3. **`tool_result` block missing its optional `content` key crashes the turn** — REPRODUCED (`KeyError` at `claude_state.py:477`).
4. **The output loop only contains `JSONDecodeError`, so any other parser exception kills the turn** — REPRODUCED (a valid-JSON-but-unhandled line raises straight out of `_process_output`).

### Assistant message with string `content` (and stray non-dict items)

**Repro steps**
1. Feed the parser an assistant line whose `message.content` is a bare string instead of a list of blocks: `{"type": "assistant", "message": {"id": "msg_str", "content": "just a plain string, not a list of blocks"}}`.
2. Call `parse_claude_code_json_lines_simple(line)` (the same parser the output loop drives via `parse_claude_code_json_lines`).

**Expected vs actual**
- Expected: the string is surfaced to the user as a single text block.
- Actual: `_handle_assistant_message` iterates the string character-by-character and evaluates `content["type"]`, raising `TypeError: string indices must be integers, not 'str'`.

**Before**
```
E   TypeError: string indices must be integers, not 'str'
sculptor/sculptor/state/claude_state.py:433: TypeError
```

**Root cause**
`_handle_assistant_message` assumed `message_data["content"]` is always a list of block dicts and iterated it directly, indexing `content["type"]` on each item (`claude_state.py:432-433`). When the CLI sends a bare string (a shape the official SDK fixed in `claude-agent-sdk-python` `d47b180` / issue #1058), iteration yields characters and `content["type"]` raises. The same code has no guard for a stray non-dict item mixed into a content list.

**Fix**
Before iterating, if `content` is a `str`, wrap it as `[{"type": "text", "text": <string>}]` so the normal text-block path runs. Inside the loop, skip any item that is not a dict so a stray non-dict element is ignored rather than indexed. This preserves the content the user should see instead of merely swallowing it.

**Test**
- Path: `sculptor/sculptor/state/claude_state_test.py` (`test_parse_assistant_message_string_content_becomes_text_block`, `test_parse_assistant_message_skips_non_dict_content_items`), `sculptor/sculptor/agents/default/claude_code_sdk/process_manager_utils_test.py` (`test_parse_lines_assistant_string_content_does_not_crash`)
- Kind: unit (criterion: buggy path is the internal CLI-stdout parser; malformed shapes are not producible via Playwright)
- Failing-test commit: `ff7d0decb84`
- Fix commit: `d7be90fc940`

**After**
```
test_parse_assistant_message_string_content_becomes_text_block PASSED
test_parse_assistant_message_skips_non_dict_content_items PASSED
test_parse_lines_assistant_string_content_does_not_crash PASSED
```

### User message with an empty content list

**Repro steps**
1. Feed the parser a user line whose content list is empty: `{"type": "user", "message": {"role": "user", "content": []}}`.
2. Call `parse_claude_code_json_lines_simple(line)`.

**Expected vs actual**
- Expected: an empty user message is skipped (parsed value `None`).
- Actual: `_handle_tool_result_message` does `message_content[0]`, raising `IndexError: list index out of range`.

**Before**
```
E   IndexError: list index out of range
sculptor/sculptor/state/claude_state.py:459: IndexError
```

**Root cause**
After the string-content check, `_handle_tool_result_message` immediately indexed `message_content[0]["type"]` (`claude_state.py:459`) with no guard for an empty list, so an empty user content list raised `IndexError` before any block could be inspected.

**Fix**
Add an explicit `if len(message_content) == 0: return None` before the first `[0]` access. `None` is already part of the function's return type and is absorbed downstream — `_load_content_for_tool_result_message` returns `None` for a `None` input, and the output loop skips `None` results — so an empty user message is simply dropped instead of crashing.

**Test**
- Path: `sculptor/sculptor/state/claude_state_test.py` (`test_parse_user_message_empty_content_list_is_skipped`), `sculptor/sculptor/agents/default/claude_code_sdk/process_manager_utils_test.py` (`test_parse_lines_empty_user_content_list_returns_none`)
- Kind: unit (same criterion as above)
- Failing-test commit: `ff7d0decb84`
- Fix commit: `d7be90fc940`

**After**
```
test_parse_user_message_empty_content_list_is_skipped PASSED
test_parse_lines_empty_user_content_list_returns_none PASSED
```

### `tool_result` block missing its `content` key

**Repro steps**
1. Feed the parser a user/tool_result line whose block omits the optional `content` key: `{"type": "user", "message": {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "toolu_missing_content", "is_error": false}]}}`.
2. Call `parse_claude_code_json_lines_simple(line)`.

**Expected vs actual**
- Expected: the tool_result parses, with the `tool_use_id` preserved and the absent content rendered as empty text.
- Actual: `tool_result["content"]` raises `KeyError: 'content'`.

**Before**
```
E   KeyError: 'content'
sculptor/sculptor/state/claude_state.py:477: KeyError
```

**Root cause**
`content` is optional on tool_result blocks in the Anthropic message format (the official SDK parses it as optional), but `_handle_tool_result_message` read it with a raising subscript `tool_result["content"]` (`claude_state.py:477`), so any tool_result that omitted the key crashed the turn.

**Fix**
Read the field via `tool_result.get("content", "")`, so a missing key yields an empty string. That value flows through unchanged: `get_tool_invocation_string` already ignores it, and the downstream `str(...)` renders empty text rather than the literal `"None"`. The block still parses with its `tool_use_id` intact.

**Test**
- Path: `sculptor/sculptor/state/claude_state_test.py` (`test_parse_tool_result_missing_content_does_not_crash`), `sculptor/sculptor/agents/default/claude_code_sdk/process_manager_utils_test.py` (`test_parse_lines_tool_result_missing_content_does_not_crash`)
- Kind: unit (same criterion as above)
- Failing-test commit: `ff7d0decb84`
- Fix commit: `d7be90fc940`

**After**
```
test_parse_tool_result_missing_content_does_not_crash PASSED
test_parse_lines_tool_result_missing_content_does_not_crash PASSED
```

### Output loop only contains `JSONDecodeError` (Layer 2 backstop)

**Repro steps**
1. Drive the real `_process_output` loop with a sequence: `system/init`, then a valid-JSON line the parser cannot handle (an assistant message missing its `message` key — deliberately a shape Layer 1 does **not** special-case, standing in for "the next unknown shape"), then a well-formed `task_notification`, then a terminating `result`.
2. Observe what escapes the loop.

**Expected vs actual**
- Expected: the bad line degrades to a `WarningAgentMessage`, and the loop keeps processing the lines after it.
- Actual: the parser's `KeyError` is not caught by `except json.JSONDecodeError`, so it propagates out of `_process_output`, ending the whole turn. The `task_notification` and `result` after it are never processed.

**Before**
```
E   KeyError: 'message'
sculptor/sculptor/state/claude_state.py:428: KeyError
(propagates out of ClaudeOutputProcessor._process_output)
```

**Root cause**
`_process_output` wrapped the `parse_claude_code_json_lines(...)` call in `except json.JSONDecodeError` only (`output_processor.py:566`). That contains non-JSON lines but not lines that are valid JSON with an unexpected shape — those raise `TypeError`/`KeyError`/`IndexError`, escape the worker thread, and kill the turn. Layer 1 removes the three known shapes, but the loop still had no backstop for the next unknown one.

**Fix**
Widen the `except` to `Exception`, routing every parse failure through the same existing warning path (`get_warning_message` + `output_message_queue.put`, line truncated to 200 chars). The warning text distinguishes a non-JSON line from an unexpected-shape line, and `get_warning_message` already logs the full exception with `logger.opt(exception=...)`. Only `Exception` is caught, so `KeyboardInterrupt`/`SystemExit` still propagate, and the `try` wraps only the parse call, so the later `AgentClientError` error-result path is unaffected.

**Test**
- Path: `sculptor/sculptor/agents/default/claude_code_sdk/output_processor_test.py` (`TestParserExceptionContainment::test_parser_exception_becomes_warning_and_loop_continues`)
- Kind: unit (drives the production `_process_output` loop verbatim)
- Failing-test commit: `00f4f9604f1`
- Fix commit: `d1da2e295e4`

**After**
```
TestParserExceptionContainment::test_parser_exception_becomes_warning_and_loop_continues PASSED
```

## Verification

- `just format`, `just check` (lint + types + ratchets), and `just test-unit` (backend, frontend, foundation, sculpt) all pass on the final tree.
- All 8 new tests fail on the pre-fix code (crashes captured above) and pass after the fix; the affected test files (`claude_state_test.py`, `process_manager_utils_test.py`, `output_processor_test.py`) run green with no regressions.

## Review notes

The Phase A4.5 self-review (`/code-review-checklist`) surfaced one finding, which was acted on in follow-up commit `5fd1a1bd733` (the Layer 2 comment referenced "Layer 1" ticket jargon a cold reader wouldn't understand; reworded to name the mechanism). Two items are recorded here rather than changed:

- **`except Exception` at `output_processor.py:566` is intentional.** This is the containment mechanism the ticket asks for. It logs the full exception, surfaces a user-visible warning (not a silent swallow), catches only `Exception` (not `BaseException`), and wraps only the parse call — so it cannot mask the later error-result control flow.
- **The Layer 2 test's malformed line relies on the parser still raising `KeyError` on a missing `message` key.** If a future change makes the parser tolerant of that shape too, the test should be re-pointed at another still-unhandled shape. This is inherent to testing a catch-all backstop.

(Sent by Claude)


trisiak:
Rebased onto `main` (merge commit `0d3a7d1`) to resolve the conflicts — the parser tests and the output-loop tests each had a neighbouring block added on both branches; both sides are kept.

Addressed all five Copilot review comments (all about comments/docstrings carrying incidental history, ticket IDs, or "Layer" jargon):

- `claude_state.py` — dropped the upstream `d47b180 / issue #1058` reference; the comment now just states the CLI can emit a bare-string `content` and we normalize it.
- `claude_state_test.py` / `process_manager_utils_test.py` — removed the `(SCU-1667)` from the section-header comments.
- `output_processor_test.py` — reworded the `TestParserExceptionContainment` docstring and inline comment to describe the behaviour directly, without "SCU-1667" or "Layer 1/2".

`just format`, `just check`, and `just test-unit` all pass on the merged tree.

(Sent by Claude)

trisiak:
## CI: fixed the two `test_error_states.py` crash failures

The offload integration run had two hard failures — `test_crash_puts_task_in_error_state` and `test_crash_shows_error_in_workspace_peek_and_persists_on_return` — both waiting for the `ERROR_INPUT` element (the ERROR-state marker) that never appeared. (The third red check, `test_pi_login_terminal_mounts_and_done_unmounts`, was flaky/unrelated — a PTY "Bad file descriptor" — and passed on retry.)

**Root cause — a real regression from this PR.** `fake_claude:crash` (the test double these two tests use) simulated an unrecoverable crash by emitting an assistant message with **string `content`** — i.e. exactly shape #1 that Layer 1 now normalizes. It relied on the resulting `TypeError` propagating out of the worker thread to push the task into ERROR state. With the hardening, that message parses cleanly (and Layer 2 would contain any parser exception anyway), so the turn no longer crashes → the task never reaches ERROR state → `ERROR_INPUT` never renders. This is the intended new behavior ("one bad message degrades to a warning, not a dead turn"); the test double was just piggybacking on the old bug.

**Fix (commit `174e12a`).** `handle_crash` now triggers the crash through a genuine state-machine invariant unrelated to parsing: it emits a `tool_result` with no assistant turn in flight, which raises `AssertionError` in `_parse_tool_result_response` (`assert self.current_turn_id is not None`). That is not an `AgentClientError`, so the agent wrapper re-raises it and the task enters ERROR state — the exact unrecoverable-vs-recoverable distinction these tests exist to check. Only the two crash tests use this command, and nothing depended on the old payload.

Validated locally via `/run-integration-test`: **2 passed in 22.3s**. Unit suites, `just check`, and `just format` remain green.

(Sent by Claude)

PR #286, opened 2026-07-05
Hold the following pin gap in both directions

DanverImbue:
## Original bug

> Failing test on main in CI — failed every run:
> `tests/integration/frontend/test_alpha_scroll_turn_end.py::test_following_keeps_pin_gap_below_last_message`

Every recent `offload integration tests` run on `main` (e.g. on eab82c54) fails with:

```
AssertionError: while following, the last message floated 89px above the viewport bottom
(expected <= 88px); the pin is parking in the paddingEnd gap
assert 89 <= (64 + 24)
```

The test guards the pin-gap contract from SCU-1673 (PR #262): while `following` a streamed
reply, the last message's bottom must sit `PIN_BOTTOM_GAP` (64px, ±24px slop) above the
viewport bottom — stated as a per-frame invariant in
`docs/development/scroll_state_unification.md`.

## Hypotheses considered

1. **Down-only pin strands the view after a mid-stream tail shrink** — REPRODUCED (root
   cause; see below).
2. **Anchoring→following handoff enters above the pin and the entry pin no-ops** —
   CONFIRMED as a secondary instance of the same asymmetry (the handoff buffer was 100px vs
   the 64px pin, so `following` could begin up to 36px above its own invariant); fixed by
   the same change plus the buffer alignment.
3. **Stability sampler races an entry transit** — DISCARDED as a test bug: the sampler
   works as designed; it faithfully reports a real plateau that should not exist.
4. **Standing measurement/DOM offset on Linux (fonts, margins, cursor line height)** —
   DISCARDED: a geometry probe on the CI-equivalent environment shows measurements,
   padding, and content height all consistent during the plateau; only `scrollTop` is off,
   and the gap converges to exactly 64px afterward.

## The bug: a mid-stream tail shrink strands the view past the pin

**Repro steps** (deterministic on CI-class hardware; verified on Offload/Modal, the same
environment CI uses):

1. Start a task with a long `fake_claude:text` reply; close the bottom panel; wait for 2
   completed messages.
2. Send a second long `fake_claude:text` follow-on; wait for 4 completed messages (history
   so the next turn genuinely anchors).
3. Send `fake_claude:stream_text` (~5k chars, 50-char chunks every 0.1s) — the reply
   overflows the anchored turn and the scroll machine hands off `anchoringTurn` →
   `following`.
4. From the moment `data-scroll-phase` flips to `following`, sample the gap between the
   last message's bottom and the viewport bottom (`get_max_following_tail_gap`).

**Expected vs actual**

- Expected: the gap is `PIN_BOTTOM_GAP` (64px) from the first `following` frame.
- Actual: the gap sits at **89px for ~500-700ms**, then converges to 64px. Probe timeline
  on Modal (~100ms windows from the phase flip; identical across runs and retries):

  ```
  before: [89, 89, 89, 89, 89, 64, 64, 64, ... 64]
  ```

  The CI test samples the gap "once stable" (two consecutive ~300ms windows within 8px) —
  on CI hardware the plateau is long enough to be certified as the resting gap → 89 > 88 →
  red. On a fast dev machine text growth usually outruns the plateau, so it passes locally
  — which is how the regression merged green.

**Before** (probe geometry on Modal during the plateau — all quantities consistent except
`scrollTop`, which is exactly one collapsed text line too deep):

```
scrollHeight − last-item DOM end = 128  (= streaming paddingEnd floor, correct)
correct pin scrollTop for DOM end 4646 = 4005
actual scrollTop                       = 4030   (stranded 25px past the pin)
gap                                    = 89     (= 64 + 25)
```

Failing test output (Offload/Modal, pre-fix — same as CI):

```
AssertionError: while following, the last message floated 89px above the viewport bottom
(expected <= 88px); the pin is parking in the paddingEnd gap
```

**Root cause**

`pinToBottom` in `useAlphaAutoScroll.ts` was **down-only** (`if (desired <= scrollTop + 1)
return`). The rule exists so a turn-end shrink (the streaming cursor unmounting) is left
in place rather than chased. But `following` promises a fixed gap every frame, and the
tail can shrink mid-stream: when the reply message first mounts it renders a standalone
blinking-cursor line, and when the first text chunk arrives that cursor collapses into the
text line — the message gets one line (~25px) shorter. The pin target moves up 25px,
down-only refuses to follow, and `scrollTop` is stranded past the pin until stream growth
overtakes the difference. The same asymmetry affected the anchoring→following handoff: it
fired while the natural anchored position was still up to 36px above the pin
(`FILLING_OVERFLOW_BUFFER` 100px vs `PIN_BOTTOM_GAP` 64px), so `following` could begin
already violating its invariant with nothing allowed to correct it.

**Fix**

Two small changes in `useAlphaAutoScroll.ts` that make the code match the one-sentence
spec — *following holds the pin gap, both directions; anchoring hands off exactly at the
pin*:

1. `pinToBottom` now applies upward corrections **only while the authority is
   `following`**, past an 8px dead-band that ignores measurement rounding. Every other pin
   site keeps the down-only rule — and the turn-end shrink stays protected, because the
   machine leaves `following` (on `streamingStopped`) before the cursor-unmount shrink
   lands, so reflows at turn end are never handled by the following pin.
2. `FILLING_OVERFLOW_BUFFER = PIN_BOTTOM_GAP`: the handoff fires exactly when the anchored
   rest position coincides with the pin target, making entry into `following` a zero-px
   event (previously the entry pin was a guaranteed no-op and the first ~36px of every
   followed stream ran above the designed gap).

Alternatives considered: fixing the cursor-collapse shrink at its source (kills this one
instance, but any future tail shrink — a markdown rewrap, a collapsing table — recreates
the class); relaxing the test band (hides a real invariant violation).

**Interplay with 437a29af** ("Harden two scroll tests against slow-runner timing"): that
commit resolved the CI symptom by widening the pin-gap band to ±40px ([24, 104]) — a
ceiling that admits exactly the one-line-stranded geometry (64 + ~25 = 89px) this bug
produces, so the guard could no longer catch its own regression. With the behavioral fix
in place the tight band is deterministically satisfiable again, so this PR restores
`_PIN_GAP_TOLERANCE_PX = 24` (upper edge deliberately below one collapsed line past the
pin). The failing-test commit is red on CI-class hardware by design; the fix commit turns
it green.

**Test**

- Path: `sculptor/tests/integration/frontend/test_alpha_scroll_turn_end.py::test_entering_following_lands_at_the_pin_gap`
- Kind: e2e (integration, Playwright)
- Failing-test commit: 73de7e14
- Fix commit: 094bc378
- Red proof (pre-fix, on Offload/Modal — the environment where the bug manifests; a fast
  dev Mac usually outruns the plateau):

  ```
  AssertionError: entering following left the last message 89px above the viewport bottom
  (expected <= 88px); the anchoring→following handoff parked above the pin instead of
  landing on it
  ```

  (`test_following_keeps_pin_gap_below_last_message` failed at 89px in the same runs —
  the same failure CI shows.)

**After** (post-fix, Offload/Modal):

- Probe timeline from following-entry: **flat `[64 ×40]`** — both attempts, plateau gone.
- All four tests in `test_alpha_scroll_turn_end.py` pass on first attempt (including the
  previously-failing CI test and the new entry-gap test).
- Locally: full frontend unit suite green; the pin-gap tests green.

```
after:  [64, 64, 64, 64, 64, 64, 64, 64, 64, ... 64]
```

## Review notes

The `/code-review-checklist` pass ran over the branch diff; acted-on findings landed as
the two follow-up commits (hook-level unit tests for the pin's direction rules, a clearer
guard structure, a docstring reword, comment trims). Outstanding items a reviewer may
want to weigh in on:

- **No embedded before/after screenshots.** The evidence slots carry the probe
  timelines, the plateau geometry, and the failing→passing test output instead: this
  workflow has no way to host binary images from the CLI (gists reject them), and the
  bug's visible surface — a transient 25px-oversized gap during streaming — is captured
  more faithfully by the numbers. Before/after screenshots from the Modal probe were
  captured and reviewed during the session and can be attached from the web UI if
  wanted.
- **Deliberate behavior change worth eyes:** a view restored at the padded max that
  re-engages `following` mid-stream now aligns to the 64px pin at the next content
  reflow (previously it stayed at the padded max for the whole stream). This is the
  documented `following` invariant; flagging since it touches the restore-adjacent
  behavior SCU-1673 preserved.
- One Offload backend-unit failure during verification was infrastructure (the sandbox
  PyPI mirror returned 500 on a download); 2059/2060 passed and `just check` covers the
  type-check locally.

(Sent by Claude)

### DIFF
--- a/sculptor/frontend/src/pages/workspace/components/chat-alpha/hooks/__tests__/useAlphaAutoScroll.test.ts
+++ b/sculptor/frontend/src/pages/workspace/components/chat-alpha/hooks/__tests__/useAlphaAutoScroll.test.ts
@@ -360,6 +360,72 @@ describe("useAlphaAutoScroll", () => {
     expect(result.current.isEngaged).toBe(true);
   });
 
+  describe("following pin direction", () => {
+    /** Engage `following` by starting a stream while the view sits at the bottom. */
+    const renderFollowing = (el: HTMLDivElement): void => {
+      const ref = { current: el };
+      const virtualizer = createMockVirtualizer();
+      const { result, rerender } = renderHook(
+        ({ isStreaming }) => useAlphaAutoScroll(ref, isStreaming, 10, virtualizer, null, -1, "test-task"),
+        { initialProps: { isStreaming: false } },
+      );
+      act(() => {
+        el.dispatchEvent(new Event("scroll"));
+      });
+      rerender({ isStreaming: true });
+      expect(result.current.isEngaged).toBe(true);
+    };
+
+    it("chases a tail shrink back up to the pin while following", () => {
+      const el = createMockScrollContainer(1300, 2000, 500); // distance=200, at bottom
+      renderFollowing(el);
+
+      // A reflow while following lands on the pin.
+      act(() => {
+        triggerResize();
+      });
+      expectPinnedToBottom(el);
+
+      // The tail shrinks by a text line with scrollTop untouched — the view is
+      // now stranded past the pin. The next reflow must correct back up.
+      setScrollPosition(el, el.scrollTop, 1975);
+      act(() => {
+        triggerResize();
+      });
+      expectPinnedToBottom(el);
+    });
+
+    it("holds a sub-dead-band overshoot instead of chasing measurement wobble", () => {
+      const el = createMockScrollContainer(1300, 2000, 500);
+      renderFollowing(el);
+      act(() => {
+        triggerResize();
+      });
+      const pinnedScrollTop = el.scrollTop;
+
+      // A shrink smaller than the dead-band: within measurement rounding — hold.
+      setScrollPosition(el, pinnedScrollTop, 1995);
+      act(() => {
+        triggerResize();
+      });
+      expect(el.scrollTop).toBe(pinnedScrollTop);
+    });
+
+    it("leaves an overshoot in place when not following (down-only preserved)", () => {
+      // Not streaming, never engaged: the authority is userControlled, and a
+      // position past the pin (inside the tail padding) is legitimate.
+      const el = createMockScrollContainer(1540, 2000, 500); // pin target is 1500
+      const ref = { current: el };
+      const virtualizer = createMockVirtualizer();
+      const { result } = renderHook(() => useAlphaAutoScroll(ref, false, 10, virtualizer, null, -1, "test-task"));
+
+      act(() => {
+        result.current.scrollToBottom();
+      });
+      expect(el.scrollTop).toBe(1540);
+    });
+  });
+
   it("scrolls to bottom when messageCount increases while at bottom (pin-to-bottom)", () => {
     // User is pinned to the bottom (distance=0), not streaming.
     // A new message arrives: messageCount increases and scrollHeight grows by 300px,
--- a/sculptor/frontend/src/pages/workspace/components/chat-alpha/hooks/useAlphaAutoScroll.ts
+++ b/sculptor/frontend/src/pages/workspace/components/chat-alpha/hooks/useAlphaAutoScroll.ts
@@ -10,7 +10,12 @@ import { useCallback, useEffect, useLayoutEffect, useRef, useState, useSyncExter
 
 import { ChatMessageRole } from "~/api";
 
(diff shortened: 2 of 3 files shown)


DanverImbue:
Done by Fable :tm:

PR #305, opened 2026-07-07
tooling: containerized-Chromium fix, iPhone-sim QA skill, portable openhost config

trisiak:
## Summary

Three small, independent dev-tooling pieces (no product code):

- **manual_test_harness**: launch Chromium with `--disable-dev-shm-usage`. Containers mount a tiny (~64MB) `/dev/shm`, and the renderer crashes with "Target/Page crashed" once it fills during a long session, especially at `device_scale_factor=2`. Redirects Chromium's shared memory to disk-backed `/tmp`.
- **auto-qa-iphone skill**: drive the web UI on a real iOS Simulator from a Mac — boots a notched iPhone, launches the local frontend server (parsing its port), taps/swipes via `idb`, and screenshots each step. Includes the Add-to-Home-Screen standalone flow, the only way to verify safe-area rendering (`env(safe-area-inset-*)` is 0 everywhere except real iOS).
- **openhost-sculptor skill**: move the per-instance deploy values (host, app, repo, branch) into a gitignored `openhost.env` sourced at startup, with a committed `openhost.env.example`, so the skill works for any instance without hardcoding a personal hostname in this public repo.

## Testing

- `ruff format --check` / `ruff check` pass on the harness change; the flag itself has been running in a containerized deployment where the long-session crashes reproduced (and stopped).
- The two skills are documentation/scripts additions; the iPhone-sim skill was exercised end-to-end on a Mac + Simulator, and the openhost skill's env sourcing on a live openhost instance.

(Sent by Claude)

### DIFF
--- a/sculptor/sculptor/testing/manual_test_harness.py
+++ b/sculptor/sculptor/testing/manual_test_harness.py
@@ -186,7 +186,15 @@ class ManualTestHarness:
         # when Chromium's GPU compositor is stubbed out, as it is in --headless=new).
         self._browser = self._playwright.chromium.launch(
             headless=True,
-            args=["--disable-webgl", "--disable-webgl2"],
+            args=[
+                "--disable-webgl",
+                "--disable-webgl2",
+                # Containers mount a tiny (~64MB) /dev/shm. Chromium's renderer uses
+                # /dev/shm for shared memory and crashes with "Target/Page crashed"
+                # once it fills during a long session (especially at device_scale_factor=2).
+                # Redirect that shared memory to disk-backed /tmp instead.
+                "--disable-dev-shm-usage",
+            ],
         )
         self._browser_context = self._browser.new_context(viewport=self._viewport, device_scale_factor=2)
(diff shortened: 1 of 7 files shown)


sculptor/sculptor/testing/manual_test_harness.py
```diff
@@ -186,7 +186,15 @@ def start(self) -> None:
...
-            args=["--disable-webgl", "--disable-webgl2"],
+            args=[
+                "--disable-webgl",
+                "--disable-webgl2",
+                # Containers mount a tiny (~64MB) /dev/shm. Chromium's renderer uses
+                # /dev/shm for shared memory and crashes with "Target/Page crashed"
+                # once it fills during a long session (especially at device_scale_factor=2).
+                # Redirect that shared memory to disk-backed /tmp instead.
+                "--disable-dev-shm-usage",
+            ],
```
trisiak: CLAUDE: this should be conditional, to avoid perf drag when running outside of shm-limitted envs

PR #345, opened 2026-07-09
[SCU-1801] Keep pending agent questions alive across Sculptor restarts

brydenfogelman:
Fixes [SCU-1801](https://linear.app/imbue/issue/SCU-1801/pending-agent-questions-lost-across-sculptor-restart).

## Problem

When an agent is blocked on an unanswered `ask_user_question` and Sculptor restarts, the pending question is lost: the interactive answer panel never re-renders (only the collapsed historical tool block remains), the task reports READY instead of WAITING, and the agent sits idle until the user re-prompts from scratch. Observed repeatedly in real usage — unanswered questions straddling restarts never came back.

## Root cause

A restart SIGTERMs the Claude CLI mid-question and the wrapper persists a `RequestStoppedAgentMessage` for the in-flight turn. Both state-derivation sites treated any stop as "the question is dead":

- `message_conversion.convert_agent_messages_to_task_update` reconstructs the pending question from the persisted `ToolUseBlock`, but the `RequestStopped` branch unconditionally cleared it — so the panel never re-rendered.
- `derived._ready_or_waiting` breaks its reverse walk on any completed request — so the task settled to READY instead of WAITING.

The question is in fact still answerable: the tool-use block is persisted, and the runner's answer-after-turn-ended continuation can deliver a late answer to a resumed session. The state was dropped purely at the derivation layer.

## Fix

**Stop attribution (`20d650a3`).** `RequestStoppedAgentMessage` gains `stopped_by_user`, recorded at emit time from the wrapper's interrupt flag. An explicit user Stop (including one that escalated past the stdin interrupt to a signal) still dismisses the question — the user is moving on. Stops the user did not ask for (backend shutdown/restart) now preserve it: `message_conversion` keeps the reconstructed question pending and `derived` keeps the task at WAITING. A newly started user turn supersedes a preserved question in both places, and so does a context clear (`bfcb9f37`) — a wiped session can no longer receive the answer. Old persisted stops lack the field and default to `False`, which matches their overwhelmingly shutdown-caused origin (a clean user Stop ends its turn via `RequestSuccess(interrupted=True)`, not `RequestStopped`); previously-lost questions therefore resurface as answerable after upgrading.

**Answer-driven relaunch settings (`5c81339f`, refactored in `d091fc61`).** Found while recording a demo of the fix: a `UserQuestionAnswerMessage` carries no model/fast-mode/effort, and those launch settings lived only in `ClaudeProcessManager` memory — so the first answer-driven turn after a restart relaunched the CLI with defaults (a fake_claude conversation crashed on the test binary stub; a real conversation silently dropped the user's fast-mode/effort). The saved message log already records every chat turn's settings and the v1 runner already replays it on task start, so the replay walk now seeds them into the wrapper via `Agent.set_conversation_launch_settings` (a documented no-op for harnesses without per-turn CLI launch state). An earlier state-file approach was replaced with this log-derived seeding to avoid a second source of truth.

## Tests

- **New e2e regression test** (`tests/integration/regression/test_regression_pending_question_restored_on_restart.py`; `9e6b278e`, strengthened in `4053a156`): drives the agent into the question-waiting state with fake_claude, SIGTERMs the backend, and asserts the question is restored end-to-end — interactive panel visible, "waiting" dot, exactly one tool block (no replay-duplication), and answering it resumes the agent to a clean completion (follow-up message lands, zero error blocks). The dot alone was proven insufficient: a failed-but-viewed task also renders a read dot, which hid a crashed resume until the assertions were strengthened.
- **New e2e launch-settings test** (`tests/integration/regression/test_regression_launch_settings_after_restart.py`; `1b3f3c2c`): sets non-default settings through the real chat controls (fast mode on, effort Low), restarts mid-question, answers — and asserts via an opt-in FakeClaude launch-args echo (`CLAUDE_FAKE_ECHO_LAUNCH_ARGS`) that the resumed turn's CLI relaunched with `fast_mode=true effort=low resumed=yes`. Verified discriminating: with the seeding gutted, it fails showing the fresh-manager defaults.
- **Derived-status unit tests** split into restart (now WAITING) and user-stop (still not WAITING) variants, plus supersession-by-new-turn, queued-but-unstarted (stays WAITING), and context-clear cases.
- **message_conversion unit tests** for preserve-vs-dismiss by stop attribution, supersession, and context clear.
- **Updated** `test_regression_replay_on_restart.py`: the restored question now legitimately pins WAITING, so the test discriminates a replay by AUQ tool-block count instead of an idle dot.

Additive `stopped_by_user` field needs no data migration (pydantic default covers old rows); only the frozen JSON-schema baseline is bumped.

## Proof of work

Failing test before the fix (panel gone after restart):

```
FAILED test_regression_pending_question_restored_on_restart.py::test_pending_question_restored_and_answerable_after_restart
AssertionError: Locator expected to be visible
  - Expect "to_be_visible" with timeout 60000ms
  - waiting for get_by_test_id("ASK_USER_QUESTION_PANEL")
```

Strengthened test before the launch-settings fix (resumed turn crashed):

```
AssertionError: Locator expected to have count '0'
Actual value: 1
  - waiting for get_by_test_id("CHAT_PANEL").get_by_test_id("ERROR_BLOCK")
```

After both fixes:

```
============================== 1 passed in 29.56s ==============================
```

- Verified against **real Claude** via the manual QA harness: question asked → backend restarted → panel restored → answered → the resumed session replied referencing the chosen answer, with no error blocks. Before/after screenshots and a screen recording of the full flow are attached to SCU-1801.
- 13 question/restart-adjacent integration files pass on Offload (64/64 tests).
- `just format` / `just check` / `just test-unit` all green.

## Review notes

- Declined: renaming `stopped_by_user` to carry an `is_`/`was_` prefix — kept for consistency with the sibling `interrupted` field on `RequestSuccessAgentMessage`.

## Deferred follow-ups

- [SCU-1802](https://linear.app/imbue/issue/SCU-1802/verify-pi-agent-pending-question-behavior-across-restart) — the pi agent has a parallel in-memory question implementation and never emits `RequestStopped`; its restart-during-question behavior is unverified and may have its own version of this bug.
- [SCU-1803](https://linear.app/imbue/issue/SCU-1803/unify-post-restart-vs-mid-lifetime-queueing-behind-a-pending-question) — post-restart, a newly typed message goes straight to the agent (dismissing the restored question by supersession) rather than queueing behind it as it would mid-lifetime. Deliberate — rebuilding the runner's waiting-set from replay risks wedged queues — but the asymmetry could be unified later.

(Sent by Claude)


DanverImbue:
Hey, I had to fix the same problem in Pi too. Perhaps there's something to unify about these two approaches? If not, it's fine, but perhaps ask Fable (while we still have it) to make sure that both Pi and Claude now have this same behavior, and correctly?

DanverImbue (approved):
LGTM!

PR #372, opened 2026-07-11
[SCU-1294] Perf: post experimental comments on every run, collapsed to a one-line summary

trisiak:
## What

Now that a `refs/notes/perf` baseline exists on `main`, make the observational perf lane comment on **every** PR run — but keep it unobtrusive by collapsing the body to a verdict + one-line summary, with the full measurement table (and its attribution) one click away inside `<details>`.

Previously the lane commented only when a PR was non-green, and rendered the full table inline.

## Changes

- **`tools/perf/comment.py`**
  - The always-visible portion is now just the verdict heading + a one-line counts summary (baseline SHA folded in). The full table and the nested attribution block move inside a `<details>`.
  - The summary is **zero-suppressed**: only non-zero buckets get a marker, so a clean run reads `✅ N unchanged · baseline abc1234` rather than leading with a red `❌ 0 regressed`.
  - The head-only (pre-baseline) renderer gets the same collapsible treatment.
  - Duplicate-row warnings stay *visible* — they flag corrupt/retried data, not a perf delta.
  - Dropped the unused `Report.is_all_green`.
- **`.github/workflows/perf.yml`**
  - Dropped the "skip comment when green" gate. The lane upserts a comment on every run (diff or head-only bootstrap), always mirrored to the step summary. The marker-based find/PATCH upsert is unchanged, so existing comments update in place.
- **`tools/perf/comment_test.py`**
  - Added tests pinning the collapsed structure (table + nested `<details>` with the blank lines GitHub requires) and the quiet all-green one-liner.

## What it looks like

Clean run (collapsed):
```
### ✅ no perf change
_Experimental — observational only; thresholds still calibrating, not a gate._
✅ 9 unchanged · baseline `46720f0`
▶ Full measurements — 9 scenario(s) (fg req · commits · dom · bg req · duration)
```
A regression flips the verdict to `⚠️ perf regressions detected` and the summary to `❌ 1 regressed · ✅ 8 unchanged · …`, with the table + attribution inside the collapsible.

## Heavy scenarios

No separate work: label-based selection already runs them. A PR labeled `performance` runs the full matrix (incl. `perf_heavy`), and `--intersect-head` keeps those rows so they diff against the full-matrix baseline note.

## Validation

`just format`, `just check`, and the perf comment unit suite (12 tests) all green. Reviewed via `/code-review-checklist` (clean — no blockers); rendered markdown eyeballed for all-green, regression, and head-only cases.

(Sent by Claude)

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -241,23 +241,19 @@ jobs:
               --base perf-out/base.jsonl --head perf-out/head.jsonl \
               --base-sha "${{ steps.baseline.outputs.base_sha }}" \
               --format github --observational --intersect-head > perf-out/comment.md
-            # Full table always goes to the step summary.
-            cat perf-out/comment.md >> "$GITHUB_STEP_SUMMARY"
-            # Comment only when something is non-green (regression/improvement/
-            # new/removed). Grep the verdict line the renderer emits.
-            if grep -q "no perf change" perf-out/comment.md; then
-              echo "should_comment=false" >> "$GITHUB_OUTPUT"
-            else
-              echo "should_comment=true" >> "$GITHUB_OUTPUT"
-            fi
           else
             # No baseline note yet: render the absolute head numbers so the PR's
             # perf profile is visible now (deltas begin once a main run seeds a
-            # note). Step summary only — nothing to diff, so no PR comment.
+            # note).
             python3 tools/perf/comment.py --head perf-out/head.jsonl \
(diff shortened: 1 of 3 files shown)


hiSandog:
Posting a compact observational result on every run seems like a good balance between visibility and noise. How will reruns or synchronize events update the existing PR comment—by a stable marker—or could they create multiple collapsed summaries? A test that runs the publisher twice for the same PR and verifies one updated comment would protect the intended unobtrusive behavior.

PR #374, opened 2026-07-13
[SCU-1827] Fix workspace rename from the context menu not focusing the input

brydenfogelman:
## What

Fixes a bug where renaming a workspace from its **sidebar row** — via the right-click context menu or the row's "..." dropdown — did not focus the inline rename input. Because the input never held focus you couldn't type, and clicking elsewhere didn't dismiss it (its `onBlur` never fired, so the rename stayed stuck open).

Fixes SCU-1827.

## Root cause

The sidebar's `beginRename` set `renamingWorkspaceIdAtom` immediately inside the menu item's `onSelect`, so the inline input mounted while the Radix menu was still closing. The menu's focus scope pulled focus back, and the freshly-mounted input's synchronous `.focus()` lost the race — leaving the input visible but unfocused.

This is the same close-time focus race SCU-1699 fixed for the panel-tab context menu and the command palette; the sidebar row menus were never converted to that deterministic handoff.

## Fix

Adopt the SCU-1699 pattern for the sidebar row menus:

- `beginRename` records intent in a new `pendingWorkspaceRenameIdAtom` instead of entering rename mode directly.
- Both row menu surfaces (`WorkspaceContextMenuContent` and the "..." `WorkspaceDropdownMenuContent`) flush that intent into `renamingWorkspaceIdAtom` from their `onCloseAutoFocus`, after the menu's focus scope has torn down — via a shared `useWorkspaceMenuCloseAutoFocus` hook. The input then mounts with nothing competing for focus.
- Corrects the `use_close_auto_focus_for_focus_handoff` review rule, which wrongly cited the workspace row menus as not handing off focus.

The recently-merged double-click rename (#355) sets `renamingWorkspaceIdAtom` directly (no overlay, no race); it coexists cleanly with this deferred menu path since both converge on the single active-rename atom.

## Tests

New Playwright integration tests in `test_workspace_row_context_menu.py`:

- rename via the context menu focuses the input — asserts `to_be_focused()`, which the prior `fill()`-based test masked (`fill` force-focuses the element)
- rename via the "..." dropdown focuses the input
- clicking elsewhere dismisses an in-progress rename
- double-click and menu rename coexist in one session (also covers the dropdown rename's click-outside dismissal)

All fail on the pre-fix code (`Locator expected to be focused`) and pass after.

## Verification

- `just check` (typecheck, lint, ratchets, file hygiene) — pass
- `just test-unit-frontend` — 3030 passed, 1 skipped
- New integration tests — pass

(Sent by Claude)

### DIFF
--- a/sculptor/frontend/src/components/CommandPalette/contextActions/atoms.ts
+++ b/sculptor/frontend/src/components/CommandPalette/contextActions/atoms.ts
@@ -19,6 +19,21 @@ import type { SubSectionId } from "~/components/sections/sectionTypes.ts";
 export const renamingWorkspaceIdAtom = atom<string | null>(null);
 
 /**
+ * A rename handoff the sidebar row menus flush once their menu has closed. The
+ * sidebar `beginRename` stashes the target workspace id here instead of writing
+ * `renamingWorkspaceIdAtom` directly: writing it while the right-click context
+ * menu or the row's "..." dropdown is still open would mount — and focus — the
+ * inline rename input inside the menu's still-active focus scope, which yanks
+ * focus back and the resulting blur cancels the rename. Each menu's
+ * `onCloseAutoFocus` is the consumer: it suppresses the focus restore, clears
+ * this atom, and writes `renamingWorkspaceIdAtom`, so the input only ever mounts
+ * with nothing competing for focus. Mirrors `palettePendingRenameAtom` for the
+ * sidebar menu surfaces; see the `use_close_auto_focus_for_focus_handoff`
+ * review rule.
+ */
+export const pendingWorkspaceRenameIdAtom = atom<string | null>(null);
+
+/**
  * A rename handoff the palette flushes once its dialog has closed. The palette
  * runtimes' `beginRename` stash the target here instead of writing the rename
  * atoms directly: writing them while the palette is open would mount — and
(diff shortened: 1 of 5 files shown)


sculptor/frontend/src/components/CommandPalette/contextActions/atoms.ts
```diff
@@ -18,6 +18,21 @@ import type { SubSectionId } from "~/components/sections/sectionTypes.ts";
...
+/**
+ * A rename handoff the sidebar row menus flush once their menu has closed. The
+ * sidebar `beginRename` stashes the target workspace id here instead of writing
+ * `renamingWorkspaceIdAtom` directly: writing it while the right-click context
+ * menu or the row's "..." dropdown is still open would mount — and focus — the
+ * inline rename input inside the menu's still-active focus scope, which yanks
+ * focus back and the resulting blur cancels the rename. Each menu's
+ * `onCloseAutoFocus` is the consumer: it suppresses the focus restore, clears
+ * this atom, and writes `renamingWorkspaceIdAtom`, so the input only ever mounts
+ * with nothing competing for focus. Mirrors `palettePendingRenameAtom` for the
```
DanverImbue, line 30: > Mirrors `palettePendingRenameAtom` for the

Comments like these go stale quite quickly. I recommend `crispy-comments`

PR #391, opened 2026-08-19
Fix OpenHost /proxy preview: use a named capture for the port (fixes 500s + phantom previews)

kilo-gi:
## Problem
On the OpenHost build, the mobile live-preview feature was broken:
- Opening any preview (`/proxy/<port>/…`) returned **HTTP 500**.
- The Previews switcher listed a **ton of broken links** — every scanned port showed up as a "live" preview.

## Root cause
`openhost-nginx.conf`'s `/proxy/<port>/` location captured the port **positionally**:
```nginx
location ~ ^/proxy/(5[1-9][0-9][0-9][0-9])(/.*)?$ {
    set $devport $1;
    ...
    proxy_pass http://127.0.0.1:$devport$request_uri;
}
```
On nginx 1.24.0, `$1` is re-resolved to garbage (`/prox`) by the time the variable-bearing `proxy_pass` is built, so the upstream becomes `127.0.0.1:/prox/proxy/<port>/` → *"invalid port in upstream"* → **500** (instead of the intended 502 → `openhost-preview-fallback.html`).

Reproduced in isolation on the same nginx:
- positional `set $devport $1` → `$devport = /prox` ❌
- named capture `(?<devport>…)` → `$devport = 51088` ✅

The switchboard (`openhost-preview-fallback.html`) treats any probe status **not** in `{502,503,504}` as a live dev server. Because every dead port returned **500** (not 502), it marked **every** scanned port as a live preview — hence the flood of broken links.

## Fix
Use a **named capture** for the port (stable through to `proxy_pass`) and drop the positional `set`. The regex is quoted because it now contains `(?<…>)`. `$request_uri` is still passed unstripped (unchanged behavior for live previews).

Dead ports now return 502 → the friendly fallback, so the switchboard lists only genuinely-live previews, and live previews load instead of 500ing.

Found during the Cloud in a Bottle catalog review.

### DIFF
--- a/openhost-nginx.conf
+++ b/openhost-nginx.conf
@@ -101,8 +101,17 @@ http {
         # NB: the port is matched as three literal [0-9] classes rather than
         # [0-9]{3} — nginx parses unquoted regex `{`/`}` as config block braces, so
         # a braced quantifier here makes nginx fail to start (pcre2 "missing )").
-        location ~ ^/proxy/(5[1-9][0-9][0-9][0-9])(/.*)?$ {
-            set $devport $1;
+        # Capture the port as a NAMED group ($devport). A positional capture
+        # ($1 via `set $devport $1`) is unreliable here: by the time the
+        # variable-bearing `proxy_pass` below is evaluated, nginx has re-resolved
+        # $1 to garbage (observed: "/prox"), producing an invalid upstream like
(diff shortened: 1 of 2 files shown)


trisiak:
LGTM, along with those sneaked in adjustments to memory

trisiak (approved):

trisiak:
You're blocked on https://github.com/imbue-ai/sculptor/pull/392 landing.

PR #439, opened 2026-09-18
Fix pi chat rendering after the 0.84 bump: message_update lost its cumulative message

DanverImbue:
## Summary

After the managed pi bump to 0.84.4, tool-call rendering in pi workspaces broke: live text streaming stopped, and a tool-loop turn's **trailing text-only assistant message vanished from the live chat** — turns ended on a bare tool chip with the agent's closing narration missing until a reload.

Root cause: pi 0.84.0 removed the cumulative `message` snapshot from `message_update` events, keeping only the `assistantMessageEvent` delta ([pi#7290](https://github.com/earendil-works/pi/issues/7290)). Sculptor's `ParsedMessageUpdate` declared `message` as required, so **every streaming update failed validation and was silently dropped** as an unknown event (verified against the real pinned binary: 40/40 `message_update` events dropped). With no `text_delta` partials, `message_conversion`'s streaming dedup treated the trailing message's lone final block as already-streamed persistence and kept only tool results.

The existing tests didn't catch it because the fixtures (`fake_pi`, the wrapper's `_text_delta_update`) still emitted the pre-0.84 wire shape — 229 tests were passing against a format pi no longer sends.

## Changes

- **`output_processor.py`** — `ParsedMessageUpdate.message` is now optional (the dispatcher only ever read the delta); model pi 0.84's new `agent_settled` event as a known-ignored variant.
- **`agent_wrapper.py`** — dispatch `agent_settled` to the ignored-events arm.
- **`fake_pi.py` / `agent_wrapper_test.py`** — update fixtures to the 0.84 delta-only shape so the harness matches the pinned wire format.
- **Regression tests** — a delta-only `message_update` parses to `ParsedMessageUpdate`; a pi tool-loop turn's trailing text renders live in `message_conversion`.

## Verification

- Replayed a real captured 0.84.4 RPC stream through the parser: every event now parses to a typed variant.
- End-to-end through the wrapper → `message_conversion`: the 0.84 stream now renders **identically** to the pre-0.84 stream (trailing text restored).
- `just format` / `just check` / `just test-unit`: all OK.
- 11 pi integration tests (background tasks, sub-agents, turn error, session resume): all pass.

## Out of scope

The `Background task completed (exit code 0).` + raw-output block Sculptor synthesizes on background-task completion is pre-existing designed behavior (pi can't self-start a reaction turn like the Claude CLI does), unchanged by this PR. It became more prominent while agent narration was being dropped; slimming it down can be a follow-up if desired.

Co-authored-by: Sculptor <sculptor@imbue.com>


DanverImbue:
image

